PR #3, opened 2017-12-31
Package nuget

gmrukwa:
First successfull approach on NuGet packaging. Current specs:
* packages NuGet package
* patches `nuspec` file version field
* uploads package to NuGet feed
* needs manual extension of `targets` file for new libs added


gmrukwa:
[Split native libraries into repositories](https://trello.com/c/G64dWsXQ/322-split-native-libraries-into-repositories)

appveyor.yml
```diff
@@ -44,19 +49,30 @@ build:
 
 after_build:
   - cmd: rmdir /s /q C:\projects\native-algorithms\src\packages
+  - cmd: 7z a cached_artifacts.zip src\Win32 src\x64 & exit 0
```
mg6: Is this `& exit 0` really necessary here? According to [this source](http://7zip.bugaco.com/7zip/MANUAL/exit_codes.htm), 7z should work fine on its own.

src/patch_nuspec.py
```diff
@@ -0,0 +1,24 @@
...
+    line_end = line[line.find(closing_tag):]
+    return line_begin + version + line_end
+
+def main():
+    version = os.getenv("APPVEYOR_BUILD_VERSION")
+    nuspec_path = 'native-algorithms.nuspec'
+    with open(nuspec_path) as nuspec:
+        content = nuspec.readlines()
+    with open(nuspec_path, 'w') as nuspec:
+        nuspec.writelines([patch_version_line(line, version) for line in content])
```
mg6: Can be simplified to just `nuspec.writelines(patch_version_line(line, version) for line in content)` (generator instead of list).

appveyor.yml
gmrukwa: During the first job, we won't have all artifacts. It returns `1` then, instead of `0`.

src/patch_nuspec.py
gmrukwa: Can be. In progress.

PR #17, opened 2018-01-24
Reenable tests

gmrukwa:
Don't know, how could this happen, but our tests became absolutely disabled at some point. This patch restores status quo.


gmrukwa:
[native-algorithms tests are not running during build](https://trello.com/c/2e8flEBv/365-native-algorithms-tests-are-not-running-during-build)

PR #13, opened 2018-01-21
Feature/contrast enhancement algorithms

danielbabiak:

PR #18, opened 2018-02-01
Feature/algorithms based on histograms

danielbabiak:


gmrukwa:
[Prepare tests for heatmap scaling algorithms and refactoring](https://trello.com/c/RamXoMUZ/353-prepare-tests-for-heatmap-scaling-algorithms-and-refactoring)

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,73 @@
...
+limitations under the License.
+*/
+
+#include "Spectre.libHeatmapDataScaling/HistogramEqualization.h"
+
+namespace spectre::visualization
+{
+std::vector<double> HistogramEqualization::scaleData(const gsl::span<double> intensities)
+{
+    std::vector<double> roundedIntensities = roundIntensities(intensities);
```
gmrukwa: If the `roundedIntensities` is not supposed to change, mark it `const`.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,73 @@
...
+    std::map<double, unsigned int> const histogramMap = calculateCumulativeDistribution(countRepeatingValues(roundedIntensities));
+    return calculateNewHistogramData(roundedIntensities, histogramMap);
+}
+
+std::vector<double> HistogramEqualization::roundIntensities(const gsl::span<double> intensities)
```
gmrukwa: This can be replaced by one line at call-site with `spectre::core::functional::transform`.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return roundedIntensities;
+}
+
+std::map<double, unsigned int> HistogramEqualization::countRepeatingValues(const gsl::span<double> intensities)
+{
+    std::map<double, unsigned int> histogramMap;
+    for (auto intensity : intensities)
+    {
+        histogramMap[intensity]++;
```
gmrukwa: We did not scale the values to fixed range before counting and rounding. Is this intended? This will work differently for the same input data just scaled by a constant factor (rounding will have proportionally different impact on the data).

BTW. I don't think we will ever need a `map`, if the data is scaled to 0-255 at the beginning. Starting from such point we may just operate on `vector` with 256 elements.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,73 @@
...
+    std::map<double, unsigned int> const histogramMap = calculateCumulativeDistribution(countRepeatingValues(roundedIntensities));
+    return calculateNewHistogramData(roundedIntensities, histogramMap);
+}
+
+std::vector<double> HistogramEqualization::roundIntensities(const gsl::span<double> intensities)
+{
+    std::vector<double> roundedIntensities(intensities.size());
+    for (int i = 0; i < intensities.size(); i++)
+    {
+        roundedIntensities[i] = round(intensities[i]);
```
gmrukwa: You're using these rounded values to map them to counts in some ranges. I'm not sure however, what will be in `roundedIntensities` after simple rounding - we still may have some slight differences, due to `double` precision, depending on scale of input data. Can't we go more stable and predictable way of realizing the same through `roundedIntensities[i] = static_cast<int>(intensities[i] + 0.5);`? @Sand3r-, I think you could validate my idea.

I am just not sure, how using `map` with `double` keys will end.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,73 @@
...
+        histogramMap[intensity]++;
+    }
+    return histogramMap;
+}
+
+std::map<double, unsigned int> HistogramEqualization::calculateCumulativeDistribution(std::map<double, unsigned int> histogramMap)
+{
+    std::map<double, unsigned int> cumulativeDistributionMap;
+    unsigned int temporaryDistributionValue = 0;
+    for (const auto singleHistogramMapElement : histogramMap)
```
gmrukwa: If you consider my earlier comments, this could be replaced by `std::accumulate`, especially when working on `std::vector`.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        temporaryDistributionValue += singleHistogramMapElement.second;
+        cumulativeDistributionMap.emplace(singleHistogramMapElement.first, temporaryDistributionValue);
+    }
+    return cumulativeDistributionMap;
+}
+
+std::vector<double> HistogramEqualization::calculateNewHistogramData(const gsl::span<double> intensities, std::map<double, unsigned int> histogramMap)
+{
+    int const size = static_cast<int>(intensities.size());
```
gmrukwa: Can `size` be negative?

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,73 @@
...
+    std::vector<double> newIntensities(size);
+    std::pair<double, unsigned int> const minCumulativeDistributionValue = (*std::min_element(histogramMap.begin(), histogramMap.end(),
+        [](decltype(histogramMap)::value_type& l, decltype(histogramMap)::value_type& r) -> bool { return l.second < r.second; }));
+    for (auto i = 0; i < size; i++)
```
gmrukwa: Consider using `transform`.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+SuppressionAlgorithm::SuppressionAlgorithm(const double topPercent) : topPercent(topPercent) { }
+
+std::vector<double> SuppressionAlgorithm::scaleData(const gsl::span<double> intensities)
+{
+    std::vector<double> newIntensities;
+    newIntensities.reserve(intensities.size());
+    const double maxIntensity = *max_element(std::begin(intensities), std::end(intensities));
+    const double cutoff = calculateQuantile(intensities, 1 - topPercent);
+    for (int i = 0; i < intensities.size(); i++)
```
gmrukwa: Consider using `transform`.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.cpp
```diff
@@ -0,0 +1,59 @@
...
+    {
+        if (intensities[i] > cutoff)
```
gmrukwa: What about extremely low values? It probably is not the case in MSI, but I am curious in general.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.cpp
```diff
@@ -0,0 +1,59 @@
...
+        else
+            newIntensities.push_back(maxIntensity * intensities[i] / cutoff);
+    }
+    return newIntensities;
+}
+
+double SuppressionAlgorithm::calculateQuantile(gsl::span<double> intensities, const double probability)
+{
+    const int size = static_cast<int>(intensities.size());
+    std::vector<double> sortedIntensities(size);
```
gmrukwa: Just initialize this as `std::vector<double> sortedIntensities(intensities.begin(), intensities.end();`.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.h
```diff
@@ -0,0 +1,50 @@
...
+    double calculateQuantile(const gsl::span<double> intensities, const double probability);
+
+    /// <summary>
+    /// Method for scaling data using suppression algorithm for contrast enhancement.
+    /// </summary>
+    /// <param name="intensities">Vector of floating point values representing intensities for each point on the heatmap.</param>
+    /// <returns>Vector of scaled floating point values representing intensities for each point on the heatmap.</returns>
+    std::vector<double> scaleData(const gsl::span<double> intensities) override;
+private:
+    const double topPercent;
```
gmrukwa: Name not conforming to convention. Should be `m_topPercent`.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.cpp
```diff
@@ -0,0 +1,59 @@
...
+    {
+        if (intensities[i] > cutoff)
+            newIntensities.push_back(cutoff);
+        else
+            newIntensities.push_back(maxIntensity * intensities[i] / cutoff);
+    }
+    return newIntensities;
+}
+
+double SuppressionAlgorithm::calculateQuantile(gsl::span<double> intensities, const double probability)
```
gmrukwa: `probability` should *probably* be renamed to `quantile`.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.cpp
```diff
@@ -0,0 +1,59 @@
...
+    std::vector<double> sortedIntensities(size);
+    for (int i = 0; i < size; i++)
+    {
+        sortedIntensities[i] = intensities[i];
+    }
+    std::sort(sortedIntensities.begin(), sortedIntensities.end());
+    const double index = 1 + (size - 1) * probability;
+    const int indexFloor = static_cast<int>(floor(index));
+    const int indexCeiling = static_cast<int>(ceil(index));
+    double quantileValue = sortedIntensities[indexFloor - 1];
```
gmrukwa: If this `-1` is due to the fact, that calculating `index`, you add 1 at the beginning, move it to calculation of `indexFloor`. Similarly with `indexCeiling`.

Ideally, you could name these magic numbers, as now it's hard to understand, from where they come.

gmrukwa (changes requested):
Although the code is probably working, I think some refactoring would be welcome. Also, I could not spot the tests.
Either way, a lot of good job done :)

EDIT: Now I see, that we're also missing tests for previous review #13. This should get fixed asap.

Sand3r-:
Please update the version in appveyor.yml file.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::vector<uint8_t> cumulativeDistribution = std::accumulate(intensitiesOccurances.begin(), intensitiesOccurances.end(), cumulativeDistribution, 
+        [](std::vector<uint8_t> accumulator, uint8_t next) {
+            if (accumulator.size() == 0)
+                accumulator.push_back(next);
+            else accumulator.push_back(accumulator[accumulator.size() - 1] + next);
+            return accumulator;
+        });
+    //return calculateNewHistogramData(roundedIntensities, histogramMap);
+    std::vector<double> returntest;
+    return returntest;
```
Daroslaw: I presume that this function is still under development?

danielbabiak: Yes, I'm working on it

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+    {
+            if (accumulator.size() == 0)
+                accumulator.push_back(intensityOccurences);
+            else accumulator.push_back(accumulator[accumulator.size() - 1] + intensityOccurences);
+            return accumulator;
+    });
+    return calculateNewHistogramData(roundedIntensities, cumulativeDistribution);
+}
+
+std::vector<uint8_t> HistogramEqualization::countRepeatingValues(const gsl::span<double> intensities)
```
Daroslaw: Any data containing values outside the range betwee 0 and 255 will result in a crash. If this method should remain public, then you should foresee a case when an unaware user feeds such data.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+*/
+
+#include "Spectre.libFunctional/Transform.h"
+#include "Spectre.libHeatmapDataScaling/HistogramEqualization.h"
+
+namespace spectre::visualization
+{
+std::vector<double> HistogramEqualization::scaleData(const gsl::span<const double> intensities)
+{
+    std::vector<double> roundedIntensities = spectre::core::functional::transform(intensities, [](double intensity) { return static_cast<int>(intensity + 0.5); });
```
Daroslaw: Is it neccessary to keep it as a vector of doubles if we round it to integer values? Maybe some `unsigned` or `unsigned long` would suffice?

src/Spectre.libHeatmapDataScaling/HistogramEqualization.h
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+#include <map>
+#include "Spectre.libHeatmapDataScaling/HeatmapDataScalingAlgorithm.h"
+
+namespace spectre::visualization
+{
+/// <summary>
+/// Class containing histogram equalization algorithm implementation.
+/// </summary>
+class HistogramEqualization : public HeatmapDataScalingAlgorithm
```
Daroslaw: The whole class relies on an implicit assumption, that values of intensity do not exceed the value 255. If this is really a thing, then I would consider putting it in some global constexpr in the header file to avoid writing magic numbers.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+                accumulator.push_back(intensityOccurences);
+            else accumulator.push_back(accumulator[accumulator.size() - 1] + intensityOccurences);
+            return accumulator;
+    });
+    return calculateNewHistogramData(roundedIntensities, cumulativeDistribution);
+}
+
+std::vector<uint8_t> HistogramEqualization::countRepeatingValues(const gsl::span<double> intensities)
+{
+    std::vector<uint8_t> histogram(256);
```
Daroslaw: You don't expect to obtain value greater than 255 for single number count? `uint8_t` seems to be a little bit too underperforming for potentially large data sets.

Daroslaw: Magic number.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+std::vector<double> HistogramEqualization::calculateNewHistogramData(const gsl::span<double> intensities, const std::vector<uint8_t> cumulativeDistribution)
+{
+    size_t const size = intensities.size();
+    std::vector<double> newIntensities(size);
+    const uint8_t minCumulativeDistributionValue = *min_element(std::begin(cumulativeDistribution), std::end(cumulativeDistribution));
+    for (size_t i = 0; i < size; i++)
+    {
+        newIntensities[i] = round(((static_cast<double>(cumulativeDistribution[static_cast<int>(intensities[i])]) - minCumulativeDistributionValue) /
+            (size - minCumulativeDistributionValue)) * 255);
```
Daroslaw: Magic number.

Daroslaw (dismissed):
I have few remarks, that potentially may be discarded because of the specifics of this algorithm. Still good job.

appveyor.yml
```diff
@@ -13,7 +13,7 @@ environment:
 configuration:
   - Debug
   - Release
-version: 4.0.0.{build}
+version: 4.1.0.{build}
```
gmrukwa: I think it should get updated now.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "Spectre.libFunctional/Transform.h"
+#include "Spectre.libHeatmapDataScaling/HistogramEqualization.h"
+
+namespace spectre::visualization
+{
+std::vector<double> HistogramEqualization::scaleData(const gsl::span<const double> intensities)
+{
+    std::vector<double> roundedIntensities = spectre::core::functional::transform(intensities, [](double intensity) { return static_cast<int>(intensity + 0.5); });
+    std::vector<uint8_t> intensitiesOccurances = countRepeatingValues(roundedIntensities);
```
gmrukwa: Typo, should be: `intensitiesOccurences`.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+    std::vector<uint8_t> cumulativeDistribution = std::accumulate(intensitiesOccurances.begin(), intensitiesOccurances.end(), cumulativeDistribution, [](std::vector<uint8_t> accumulator, uint8_t intensityOccurences)
```
gmrukwa: What is the behavior of using not initialized `cumulativeDistribution` object here? I had something like [this](https://ideone.com/5d29C1). If you want empty `std::vector<uint8_t>` just put `std::vector<uint8_t>()`.

On the other hand, you should pre-allocate the vector, since you know the final size it will have.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+            return accumulator;
+    });
+    return calculateNewHistogramData(roundedIntensities, cumulativeDistribution);
+}
+
+std::vector<uint8_t> HistogramEqualization::countRepeatingValues(const gsl::span<double> intensities)
+{
+    std::vector<uint8_t> histogram(256);
+    for (int i = 0; i < intensities.size(); ++i) {
+        histogram[static_cast<uint8_t>(intensities[i])]++;
```
gmrukwa: Use pre-incrementation, since you're not assigning the old value. Otherwise a copy of this value is made.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+    }
+    return histogram;
+}
+
+std::vector<double> HistogramEqualization::calculateNewHistogramData(const gsl::span<double> intensities, const std::vector<uint8_t> cumulativeDistribution)
```
gmrukwa: - `gsl::span<const double> intensities`
- `cumulativeDistribution` should either be passed as `gsl::span` or by `const` reference. Since it's the internal function in your object, I'd go for reference.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+    size_t const size = intensities.size();
+    std::vector<double> newIntensities(size);
+    const uint8_t minCumulativeDistributionValue = *min_element(std::begin(cumulativeDistribution), std::end(cumulativeDistribution));
+    for (size_t i = 0; i < size; i++)
+    {
+        newIntensities[i] = round(((static_cast<double>(cumulativeDistribution[static_cast<int>(intensities[i])]) - minCumulativeDistributionValue) /
```
gmrukwa: Separate this into meaningful expressions or extract as an `inline` function that does the job. It's too long.

src/Spectre.libHeatmapDataScaling/ScalingDataUtils.h
```diff
@@ -0,0 +1,29 @@
...
+#include <vector>
+
+namespace spectre::visualization
+{
+/// <summary>
+/// Method for data normalization to the given ranges.
+/// </summary>
+/// <param name="intensities">Vector of floating point values representing data for scaling.</param>
+/// <returns>Vector of normalized floating point values representing scaled data.</returns>
+std::vector<double> normalizeData(const gsl::span<double> intensities, const int minIntensityRange = 0, const int maxIntensityRange = 255);
```
gmrukwa: This is just scaling. It should be named appropriately.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.h
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+    /// <summary>
+    /// Method for scaling data using histogram equalization algorithm for contrast enhancement.
+    /// </summary>
+    /// <param name="intensities">Vector of floating point values representing intensities for each point on the heatmap.</param>
+    /// <returns>Vector of scaled floating point values representing intensities for each point on the heatmap.</returns>
+    std::vector<double> scaleData(const gsl::span<const double> intensities) override;
```
gmrukwa: It should be marked as `const`, since it does not change the state of `HistogramEqualization` instance.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.h
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Method for counting repeating values in the vector.
+    /// </summary>
+    /// <param name="intensities">Vector of floating point values of intensities.</param>
+    /// <returns>Vector which stores the number of occurrences of the intensity representing by index.</returns>
+    std::vector<uint8_t> countRepeatingValues(const gsl::span<double> intensities);
```
gmrukwa: It should be marked as `const`, since it does not change the state of `HistogramEqualization` instance.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.h
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Method for calculating new histogram data based on input vector of intensities and cumulative distribution.
+    /// </summary>
+    /// <param name="intensities">Vector of floating point values of intensities.</param>
+    /// <param name="histogramMap">Map containing intensity value as a key and cumulative distribution as a value.</param>
+    /// <returns>Vector of scaled floating point values representing intensities for each point on the heatmap.</returns>
+    std::vector<double> calculateNewHistogramData(const gsl::span<double> intensities, const std::vector<uint8_t> cumulativeDistribution);
```
gmrukwa: It should be marked as `const`, since it does not change the state of `HistogramEqualization` instance.

src/Spectre.libHeatmapDataScaling/HeatmapDataScalingAlgorithm.h
```diff
@@ -36,6 +36,6 @@ class HeatmapDataScalingAlgorithm
     /// </summary>
     /// <param name="intensities">Vector of floating point values representing intensities for each point on the heatmap.</param>
     /// <returns>Vector of scaled floating point values representing intensities for each point on the heatmap.</returns>
-    virtual std::vector<double> scaleData(const gsl::span<double> intensities) = 0;
+    virtual std::vector<double> scaleData(const gsl::span<const double> intensities) = 0;
```
gmrukwa: It should be marked as `const`, since it does not change the state of `HeatmapDataScalingAlgorithm` instance.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "Spectre.libHeatmapDataScaling/SuppressionAlgorithm.h"
+
+namespace spectre::visualization
+{
+SuppressionAlgorithm::SuppressionAlgorithm(const double topPercent) : m_topPercent(topPercent) { }
+
+std::vector<double> SuppressionAlgorithm::scaleData(const gsl::span<const double> intensities)
+{
+    const double maxIntensity = *max_element(std::begin(intensities), std::end(intensities));
+    const double cutoff = calculateQuantile(intensities, 1 - m_topPercent);
```
gmrukwa: If this `1 - m_topPercent` passes, you have *warning level* set to 3 instead of 4, or disabled *treating warnings as errors*.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.h
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    /// <param name="topPercent">Percent of outliers which will be suppressed.</param>
+    SuppressionAlgorithm(const double topPercent = 0.01);
+
+    /// <summary>
+    /// Method for calculating quantile.
+    /// </summary>
+    /// <param name="probability">Determines the quantile.</param>
+    /// <returns>Quantile value.</returns>
+    double calculateQuantile(const gsl::span<const double> intensities, const double probability);
```
gmrukwa: It should be marked as `const`, since it does not change the state of `SuppressionAlgorithm` instance.

src/Spectre.libHeatmapDataScaling/SuppressionAlgorithm.h
```diff
@@ -0,0 +1,50 @@
...
+
+    /// <summary>
+    /// Method for scaling data using suppression algorithm for contrast enhancement.
+    /// </summary>
+    /// <param name="intensities">Vector of floating point values representing intensities for each point on the heatmap.</param>
+    /// <returns>Vector of scaled floating point values representing intensities for each point on the heatmap.</returns>
+    std::vector<double> scaleData(const gsl::span<const double> intensities) override;
```
gmrukwa: It should be marked as `const`, since it does not change the state of `SuppressionAlgorithm` instance.

gmrukwa (dismissed):

appveyor.yml
```diff
@@ -13,7 +13,7 @@ environment:
 configuration:
   - Debug
   - Release
-version: 4.1.0.{build}
+version: 4.3.0.{build}
```
gmrukwa: I am pretty sure, we have to update it even once more. Current version is `4.4.0`. Now it will be `4.5.0`.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -14,43 +14,55 @@ See the License for the specific language governing permissions and
...
+		throw core::exception::ArgumentOutOfRangeException<double>("minIntensitiesRange", 0, 255, minIntensitiesRange);
+	}
+	if (maxIntensitiesRange > 255)
+	{
+		throw core::exception::ArgumentOutOfRangeException<double>("maxIntensitiesRange", 0, 255, maxIntensitiesRange);
+	}
+
+	std::vector<uint8_t> roundedIntensities = spectre::core::functional::transform<double, uint8_t>(intensities, [](double intensity) { return static_cast<uint8_t>(intensity + 0.5); });
+    std::vector<unsigned int> intensitiesOccurences = countRepeatingValues(roundedIntensities);
+	const std::vector<unsigned int> cumulativeDistribution = std::accumulate(intensitiesOccurences.begin(), intensitiesOccurences.end(), std::vector<unsigned int>(), [](std::vector<unsigned int> accumulator, unsigned int intensityOccurences)
```
gmrukwa: You are copying `accumulator` in each call to the lambda. This is performance issue.

I digged a bit. It appears that [`std::partial_sum`](http://en.cppreference.com/w/cpp/algorithm/partial_sum) will be **much** more fitting this purpose than `std::accumulate`. It will be a pleasant one-liner, without this error-prone lambda below. 😃

src/Spectre.libHeatmapDataScaling/HistogramEqualization.h
```diff
@@ -20,6 +20,9 @@ limitations under the License.
 
 namespace spectre::visualization
 {
+
+constexpr uint8_t intensitiesRange = 255;
```
gmrukwa: I am not sure we want to pollute the namespace `spectre::visualization` with a value so *specific* for this algorithm. I'd rather put this close to the place it is used: `HistogramEqualization.cpp`, probably at the top of the file.

I also think you should use [`std::numeric_limits<uint8_t>::max()`](http://en.cppreference.com/w/cpp/types/numeric_limits/max) instead.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -14,43 +14,55 @@ See the License for the specific language governing permissions and
...
-    std::vector<double> roundedIntensities = spectre::core::functional::transform(intensities, [](double intensity) { return static_cast<int>(intensity + 0.5); });
-    std::vector<uint8_t> intensitiesOccurances = countRepeatingValues(roundedIntensities);
-    std::vector<uint8_t> cumulativeDistribution = std::accumulate(intensitiesOccurances.begin(), intensitiesOccurances.end(), cumulativeDistribution, [](std::vector<uint8_t> accumulator, uint8_t intensityOccurences)
+	const auto minIntensitiesRange = *min_element(std::begin(intensities), std::end(intensities));
+	const auto maxIntensitiesRange = *max_element(std::begin(intensities), std::end(intensities));
+	if (minIntensitiesRange < 0)
+	{
+		throw core::exception::ArgumentOutOfRangeException<double>("minIntensitiesRange", 0, 255, minIntensitiesRange);
+	}
+	if (maxIntensitiesRange > 255)
```
gmrukwa: Magic number.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -14,43 +14,55 @@ See the License for the specific language governing permissions and
...
+std::vector<double> HistogramEqualization::scaleData(const gsl::span<const double> intensities) const
 {
-    std::vector<double> roundedIntensities = spectre::core::functional::transform(intensities, [](double intensity) { return static_cast<int>(intensity + 0.5); });
-    std::vector<uint8_t> intensitiesOccurances = countRepeatingValues(roundedIntensities);
-    std::vector<uint8_t> cumulativeDistribution = std::accumulate(intensitiesOccurances.begin(), intensitiesOccurances.end(), cumulativeDistribution, [](std::vector<uint8_t> accumulator, uint8_t intensityOccurences)
+	const auto minIntensitiesRange = *min_element(std::begin(intensities), std::end(intensities));
+	const auto maxIntensitiesRange = *max_element(std::begin(intensities), std::end(intensities));
+	if (minIntensitiesRange < 0)
+	{
+		throw core::exception::ArgumentOutOfRangeException<double>("minIntensitiesRange", 0, 255, minIntensitiesRange);
```
gmrukwa: Magic number.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -14,43 +14,55 @@ See the License for the specific language governing permissions and
...
+	}
+	if (maxIntensitiesRange > 255)
+	{
+		throw core::exception::ArgumentOutOfRangeException<double>("maxIntensitiesRange", 0, 255, maxIntensitiesRange);
```
gmrukwa: Magic number.

src/Spectre.libHeatmapDataScaling/HistogramEqualization.cpp
```diff
@@ -14,43 +14,55 @@ See the License for the specific language governing permissions and
...
 }
 
-std::vector<uint8_t> HistogramEqualization::countRepeatingValues(const gsl::span<double> intensities)
+std::vector<unsigned int> HistogramEqualization::countRepeatingValues(const gsl::span<uint8_t> intensities) const
 {
-    std::vector<uint8_t> histogram(256);
+    std::vector<unsigned int> histogram(intensitiesRange + 1);
     for (int i = 0; i < intensities.size(); ++i) {
-        histogram[static_cast<uint8_t>(intensities[i])]++;
+        ++histogram[static_cast<unsigned long>(intensities[i])];
```
gmrukwa: `static_cast<size_t>` maybe?

src/Spectre.libHeatmapDataScaling/ScalingDataUtils.h
```diff
@@ -21,9 +21,9 @@ limitations under the License.
...
 {
 /// <summary>
-/// Method for data normalization to the given ranges.
+/// Method for data scaling data to the given ranges.
 /// </summary>
 /// <param name="intensities">Vector of floating point values representing data for scaling.</param>
-/// <returns>Vector of normalized floating point values representing scaled data.</returns>
-std::vector<double> normalizeData(const gsl::span<double> intensities, const int minIntensityRange = 0, const int maxIntensityRange = 255);
+/// <returns>Vector of floating point values representing scaled data.</returns>
+std::vector<double> scaleDataToRange(const gsl::span<double> intensities, const int minIntensityRange = 0, const int maxIntensityRange = 255);
```
gmrukwa: Use [`std::numeric_limits<uint8_t>::max()`](http://en.cppreference.com/w/cpp/types/numeric_limits/max).

gmrukwa (dismissed):
Few more clarity issues and one performance issue.

gmrukwa (dismissed):

Daroslaw (dismissed):

mg6:
@danielbabiak Please incorporate changes from `develop` and add `CMakeLists.txt` file for Linux build. Have a look on how it’s done here:
- https://github.com/spectre-team/native-algorithms/blob/develop/src/Spectre.libPeakFinder.Tests/CMakeLists.txt
- https://github.com/spectre-team/native-algorithms/blob/develop/CMakeLists.txt#L71

Daroslaw (approved):

gmrukwa (approved):

PR #20, opened 2018-02-05
Feature - wavelet denoising - wavelet decomposer

Sand3r-:
First and foremost, commits introduce the reference version of wavelet decomposer, backed up with numerous comments and references. Secondly, the convolution class is modified so that it can make use of the format of precomputed Daubechies 4 filters. Lastly, the filter coefficients themselves have got their names revised as well.

### DIFF
--- a/src/Spectre.libWavelet.Tests/ConvolutionTest.cpp
+++ b/src/Spectre.libWavelet.Tests/ConvolutionTest.cpp
@@ -19,6 +19,7 @@ limitations under the License.
 
 #include <gtest/gtest.h>
 #include "Spectre.libWavelet\Convolution.h"
+#include "Spectre.libFunctional\Range.h"
 
 namespace
 {
@@ -31,22 +32,13 @@ namespace
 
     class ConvolutionTest : public ::testing::Test
     {
-    public:
-        void InitializeSignal(Signal& signal)
-        {
-            for (unsigned i = 0; i < signal.size(); i++)
-            {
-                signal[i] = (DataType)i;
-            }
-        }
     protected:
         Convolution convolution;
     };
 
     TEST_F(ConvolutionTest, properly_convolves_the_kernel_over_a_signal)
     {
-        Signal signal(10);
-        InitializeSignal(signal);
+        Signal signal = spectre::core::functional::range<DataType>(10);
         Kernel kernel = { 0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0 };
         Signal result = convolution.Convolve(kernel, signal);
         Signal correctResult = { 0.0, 0.0, 1.0, 4.0, 10.0, 20.0, 35.0, 56.0, 84.0, 112.0 };
--- a/src/Spectre.libWavelet.Tests/ConvolutionTest.cpp
+++ b/src/Spectre.libWavelet.Tests/ConvolutionTest.cpp
@@ -31,16 +31,25 @@ namespace
 
     class ConvolutionTest : public ::testing::Test
     {
+    public:
+        void InitializeSignal(Signal& signal)
+        {
+            for (unsigned i = 0; i < signal.size(); i++)
+            {
+                signal[i] = (DataType)i;
+            }
+        }
     protected:
         Convolution convolution;
     };
 
     TEST_F(ConvolutionTest, properly_convolves_the_kernel_over_a_signal)
     {
-        Signal signal = { 1.0, 2.0, 3.0, 4.0 };
-        Signal kernel = { -0.5, 1.0 };
+        Signal signal(10);
+        InitializeSignal(signal);
+        Kernel kernel = { 0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0 };
         Signal result = convolution.Convolve(kernel, signal);
-        Signal correctResult = { -0.5, 0.0, 0.5, 1.0 };
+        Signal correctResult = { 0.0, 0.0, 1.0, 4.0, 10.0, 20.0, 35.0, 56.0, 84.0, 112.0 };
         ASSERT_EQ(correctResult, result);
     }
 }
--- a/src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
+++ b/src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
@@ -21,20 +21,12 @@ limitations under the License.
 #include <gtest/gtest.h>
 #include "Spectre.libWavelet\WaveletDecomposerRef.h"
 #include "Spectre.libFunctional\Range.h"
+#include <gmock/gmock-matchers.h>
+#include <gmock/gmock-more-actions.h>
 
 namespace
 {
 using namespace spectre::algorithm::wavelet;
-
-static void CompareVectors(const std::vector<DataType>& estimate,
-    const std::vector<DataType>& correct)
-{
-    constexpr DataType maxAbsoluteError = 0.0001;
-    for (unsigned i = 0; i < estimate.size(); i++)
-    {
-        ASSERT_NEAR(estimate[i], correct[i], maxAbsoluteError);
-    }
-}
 
 TEST(WaveletDecomposerInitialization, initializes)
 {
@@ -114,8 +106,7 @@ public:
                 1.23299325679310,
                 1.41702876032074,
                 0.461966379081157,
-                3.55408918941385e-05,
-                0.0
+                3.55408918941385e-05
             };
         lastLowFrequencyResult = {
                 -4.64283404088808e-07,
@@ -132,7 +123,6 @@ public:
                 1.41934784856588,
                 0.463535004863161,
                 6.73843951653069e-05,
-                0.0,
                 0.0
             };
     }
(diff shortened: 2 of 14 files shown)


src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
```diff
@@ -0,0 +1,171 @@
...
+        std::vector<DataType> lastLowFrequencyResult;
+    };
+
+    TEST_F(WaveletDecomposerRefTest, decomposes_the_signal)
+    {
+        Signal signal(10);
+        InitializeSignal(signal);
+        WaveletCoefficients coefficients = decomposer.Decompose(std::move(signal));
+
+        // Due to high amount of coefficients generated, only a few sets will be
```
Regis128: Again unnecessary split when code lines are wider than unsplit comment would be, makes it less transparent

src/Spectre.libWavelet/PrecomputedDaubechiesCoefficients.h
```diff
@@ -1,5 +1,5 @@
 /*
- * PrecomputedDaubechiesCoefficients.h
+ * precomputedDaubechies4Coefficientss.h
```
Regis128: Class name typo

src/Spectre.libWavelet/WaveletDecomposerRef.h
```diff
@@ -0,0 +1,59 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+#pragma once
+#include "Convolution.h"
+#include "DataTypes.h"
+
+namespace spectre::algorithm::wavelet
+{
+    constexpr unsigned int WAVELET_LEVELS = 10;
```
Regis128: again, indent to fix

src/Spectre.libWavelet/WaveletDecomposerRef.h
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Decomposes the signal into set of Daubechies coefficients.
+    /// </summary>
+    class WaveletDecomposerRef
+    {
+    public:
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WaveletDecomposer"/> class.
+        /// </summary>
+        /// <param name="threshold">Treshold to be used.</param>
```
Regis128: there is no param

src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
```diff
@@ -0,0 +1,171 @@
+/*
+* WaveletDecomposerRefTest.cpp
+* Tests whether the signal is properly decomposed by the reference
```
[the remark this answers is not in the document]
Sand3r-: I'm having trouble understanding. What docs have been splitted?

src/Spectre.libWavelet/WaveletDecomposerRef.h
Sand3r-: Nice catch!

src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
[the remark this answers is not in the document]
Sand3r-: I have adjusted the code to take up to 83 characters on each line, so that it conforms to the coding standards rule.

src/Spectre.libWavelet/WaveletDecomposerRef.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "PrecomputedDaubechiesCoefficients.h"
+
+namespace spectre::algorithm::wavelet
+{
+static constexpr size_t ComputeScale(unsigned level)
+{
+    return 1ULL << (level + 1); // 2^(L + 1)
+}
+
+static constexpr size_t CeiledDivision(size_t dividend, size_t divisor)
```
Daroslaw: I'm not sure if the `constexpr` really does anything here. There is no invocation involving constants that could be evaluated in the compile time (https://stackoverflow.com/a/14248310) and I doubt that the finite amount of combination of those two arguments is so low that the compiler makes up a look-up table (of course if such scenario is even a thing). The `ComputeScale` function could also be a questionable case, but if compiler goes deep into the semantics, it could find out the `WAVELET_LEVELS` and actually pre-compute those few values.

Sand3r-: Well, apparently it optimizes it out.
image

Sand3r-: It actually doesn't precompute the value.
image

src/Spectre.libWavelet.Tests/ConvolutionTest.cpp
```diff
@@ -31,16 +31,25 @@ namespace
 
     class ConvolutionTest : public ::testing::Test
     {
+    public:
+        void InitializeSignal(Signal& signal)
```
gmrukwa: Can't we just go for producing this signal inside and returning it?

On the other hand, isn't it just an equivalent of calling `spectre::core::functional::range<DataType>(signal.size())`?

src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
```diff
@@ -0,0 +1,175 @@
...
+*/
+
+#include <gtest/gtest.h>
+#include "Spectre.libWavelet\WaveletDecomposerRef.h"
+
+namespace
+{
+using namespace spectre::algorithm::wavelet;
+
+static void CompareVectors(const std::vector<DataType>& estimate,
```
gmrukwa: Isn't that already in GTest? Check [this](https://stackoverflow.com/a/2797990/3067956) and [this](https://github.com/google/googletest/blob/master/googlemock/docs/CheatSheet.md#floating-point-matchers). It should also provide better reporting of failure.

src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
```diff
@@ -0,0 +1,175 @@
...
+                1.23076472595957,
+                1.41934784856588,
+                0.463535004863161,
+                6.73843951653069e-05,
+                0.0,
+                0.0
+            };
+    }
+
+    void InitializeSignal(Signal& signal)
```
gmrukwa: `range`?

src/Spectre.libWavelet/Convolution.cpp
```diff
@@ -24,7 +24,7 @@ Convolution::Convolution()
 {
 }
 
-Signal Convolution::Convolve(const Signal& kernel, const Signal &signal)
+Signal Convolution::Convolve(Kernel& kernel, const Signal &signal) const
```
gmrukwa: UMMM, GOD IT'S BEAUTIFUL NOW. :)

src/Spectre.libWavelet/Convolution.h
```diff
@@ -17,11 +17,14 @@
    limitations under the License.
 */
 #pragma once
+#include <array>
 #include <span.h>
 #include "DataTypes.h"
 
 namespace spectre::algorithm::wavelet
 {
+using Kernel = const std::array<const DataType, 8>;
```
gmrukwa: Maybe we could remove first `const` from here...

src/Spectre.libWavelet/Convolution.h
```diff
@@ -39,6 +42,6 @@ class Convolution
     /// <param name="kernel">Kernel to be used.</param>
     /// <param name="signal">Signal to be convolved.</param>
     /// <returns>Filtered signal.</returns>
-    Signal Convolve(const Signal& kernel, const Signal& signal);
+    Signal Convolve(Kernel& kernel, const Signal& signal) const;
```
gmrukwa: ... and place it here? (`const Kernel& kernel`)? I just thought that signature would be more explicit on IDE hints.

src/Spectre.libWavelet/PrecomputedDaubechiesCoefficients.h
```diff
@@ -32,44 +32,44 @@ constexpr std::array<const DataType, 8> precomputedDaubechiesCoefficient{
...
-constexpr std::array<const DataType, 8> ReconstructionVlNumerator{
-    precomputedDaubechiesCoefficient[0],
-    precomputedDaubechiesCoefficient[1],
-    precomputedDaubechiesCoefficient[2],
-    -precomputedDaubechiesCoefficient[3],
-    -precomputedDaubechiesCoefficient[4],
-    precomputedDaubechiesCoefficient[5],
-    precomputedDaubechiesCoefficient[6],
-    -precomputedDaubechiesCoefficient[7]
+constexpr std::array<const DataType, 8> ReconstructionLowPassFilter{
```
gmrukwa: Insightful analysis. Good job!

src/Spectre.libWavelet/WaveletDecomposerRef.cpp
```diff
@@ -0,0 +1,126 @@
...
+    {
+        lowFrequencyCoefficients[i].resize(blockLength); // Extend with zeros/shrink to match size
+        coefficients.data[level][i] = // Apply Daubechies4 decomposition high pass filter
+            m_Convolution.Convolve(precomputed::DecompositionHighPassFilter, lowFrequencyCoefficients[i]);
+        lowFrequencyCoefficients[i] = // Apply Daubechies4 decomposition low pass filter
+            m_Convolution.Convolve(precomputed::DecompositionLowPassFilter, lowFrequencyCoefficients[i]);
+    }
+}
+
+WaveletDecomposerRef::WaveletDecomposerRef() : m_Convolution()
```
gmrukwa: And in such a case I wonder, whether `Convolution` should be a `class` indeed. Class does no partial application. But let's leave it this way now.

src/Spectre.libWavelet/WaveletDecomposerRef.cpp
Sand3r-: Btw. found some useful bitshifts out there:
image

src/Spectre.libWavelet/WaveletDecomposerRef.cpp
```diff
@@ -0,0 +1,126 @@
...
+        scale = ComputeScale(currentLevel);
+        blockLength = ComputeBlockLength(currentLevel + 1, scale, signalLength);
+        DownsampleByDissolving(lowFrequencyCoefficients, currentLevel + 1, (unsigned)blockLength);
+    }
+
+    ApplyFilters(coefficients, lowFrequencyCoefficients, scale, blockLength, WAVELET_LEVELS - 1);
+
+    return coefficients;
+}
+}
```
gmrukwa: This file consists mostly of `static inline` functions, that are inivisble anywhere else outside of this file. I know, it's used nowhere else, but limits our capabilities of testing. Let's leave it this way for now, but I suppose you'll need to make this more testable (and to implement tests) before you'll start speeding this up. Just for checking consistence other way than reading values of variables in *Debug*.

src/Spectre.libWavelet/WaveletDecomposerRef.h
```diff
@@ -0,0 +1,58 @@
...
+    explicit WaveletDecomposerRef();
+    /// <summary>
+    /// Decomposes the signal into wavelet coefficients.
+    /// </summary>
+    /// <param name="signal">Signal to decompose.</param>
+    /// <returns>Set of Daubechies wavelet coefficients.</returns>
+    WaveletCoefficients Decompose(Signal&& signal) const;
+
+private:
+    inline void WaveletDecomposerRef::ApplyFilters(
```
gmrukwa: Hmmm... I see a little bit different convention as compared to the one we were using until now (i.e. check `libGenetic`). In general, we try to consider a function to be *function* in mathematical sense, mapping between input and output. I know, we may consider `coefficients` and `lowFrequencyCoefficients` to be outputs, but its calling convention is different than we have in most of the code.

I'll accept any way we choose - shame, you weren't available when `libGenetic` was built. We could agree on a single version back then. Especially, that I see, that your style is much more popular in general. Either way, I think consistency (at least to ourselves) is one of the crucial points we have to keep.

src/Spectre.libWavelet/WaveletDecomposerRef.cpp
Sand3r-: Yeah well, I do believe that the standard way of hiding the implementation details from the public interface is creating 2 classes, for example:
Denoiser
DenoiserImpl

Where Denoiser has a single public method Denoise which calls the DenoiserImpl method(s).
...And DenoiserImpl has all those actual methods implemented as public for the sake of testing.

The reason I have made them inlines/static was done for the sake of achieving a bit of a speed up during compile and run times.

src/Spectre.libWavelet/WaveletDecomposerRef.h
Sand3r-: Oh well, I didn't know that. I made it in such a way to avoid data copies, and keep the last bits of cache locality. All there is to it, is that it's just more efficient. I do understand that the it may feel more natural to have it the math-function-resembling manner, but after all, is it a necessity that algorithms that do not interleave need to have the same convention?

src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
Sand3r-: I do not really see a way to achieve that with the default functionality.
In this [answer](https://stackoverflow.com/a/13112240) the author actually implements his own matcher for floating point array comparison

src/Spectre.libWavelet/WaveletDecomposerRef.h
gmrukwa: Ok, fine - but we can always `std::move` the data, can't we? :)

Sand3r-: ...Which obfuscates the code, yes we do. And which mostly works on heap-allocated data, and sometimes (depending on scope) in stack.

src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
gmrukwa: As we spoke, I suppose it will be implemented with my matcher chunk into this review?

src/Spectre.libWavelet/WaveletDecomposerRef.cpp
gmrukwa: I just wasn't sure at the beginning. Thought we'll use class to provide some arguments once.

src/Spectre.libWavelet/WaveletDecomposerRef.h
gmrukwa: As we spoke outside, I'll accept this since it affects 2 `private inline` functions and API does not get inconsistent.

src/Spectre.libWavelet.Tests/WaveletDecomposerRefTest.cpp
Sand3r-: It will

PR #25, opened 2018-02-23
Soft tresholder for signal denoising

Sand3r-:
Soft tresholder filters out the coefficients of decomposed signal.

The MeanAbsoluteDeviationNoiseEstimator has been revamped to Median(...) to comply with original version.

### DIFF
--- a/src/Spectre.libStatistics.Tests/StatisticsTest.cpp
+++ b/src/Spectre.libStatistics.Tests/StatisticsTest.cpp
@@ -58,9 +58,9 @@ TEST(MedianTest, calculates_vector_median)
     EXPECT_THAT(Median(data), 2.);
 }
 
-TEST(MedianTest, throws_on_empty_vector)
+TEST(MedianTest, median_of_empty_is_zero)
 {
-    EXPECT_THROW(MedianAbsoluteDeviation(empty), spectre::core::exception::EmptyArgumentException);
+    EXPECT_THAT(Median(empty), 0.);
 }
 
 TEST(VarianceTest, calculates_unbiased_vector_variance_by_default)
@@ -110,8 +110,8 @@ TEST(MedianAbsoluteDeviationTest, calculates_median_absolute_deviation_of_vector
     EXPECT_THAT(MedianAbsoluteDeviation(data), DoubleEq(1.0));
 }
 
-TEST(MedianAbsoluteDeviationTest, throws_exception_when_set_is_empty)
+TEST(MedianAbsoluteDeviationTest, median_absolute_deviation_of_empty_is_zero)
 {
-    EXPECT_THROW(MedianAbsoluteDeviation(empty), spectre::core::exception::EmptyArgumentException);
+    EXPECT_THAT(MedianAbsoluteDeviation(empty), 0.);
 }
 }
--- a/src/Spectre.libStatistics.Tests/StatisticsTest.cpp
+++ b/src/Spectre.libStatistics.Tests/StatisticsTest.cpp
@@ -2,7 +2,7 @@
 * StatisticsTest.cpp
 * Tests basic vector statistics.
 *
-Copyright 2017 Grzegorz Mrukwa
+Copyright 2018 Grzegorz Mrukwa, Michal Gallus
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


Sand3r-:
[Implement wavelet denoising](https://trello.com/c/KzPr7RPc/205-implement-wavelet-denoising)

src/Spectre.libStatistics/Statistics.h
```diff
@@ -48,6 +50,22 @@ constexpr DataType Mean(gsl::span<const DataType> data)
...
+/// </summary>
+/// <param name="data">The data.</param>
+/// <returns>Median of the elements.</returns>
+template <class DataType>
+DataType Median(gsl::span<const DataType> data)
+{
+    static_assert(std::is_arithmetic_v<DataType>, "DataType: expected arithmetic.");
+    std::vector<std::remove_const<DataType>::type> sortedData(data.begin(), data.end());
+    std::sort(sortedData.begin(), sortedData.end());
+    if (data.size() == 0) throw spectre::core::exception::EmptyArgumentException("Median: Provided vector is empty");
```
Daroslaw: How about checking this before sorting operation?

Daroslaw (dismissed):
Good job overall!

src/Spectre.libStatistics/Statistics.h
Sand3r-: Good point actually :smile:

src/Spectre.libWavelet/MedianAbsoluteDeviationNoiseEstimator.cpp
```diff
@@ -17,21 +17,21 @@
...
+    MedianAbsoluteDeviationNoiseEstimator::MedianAbsoluteDeviationNoiseEstimator(DataType multiplier) :
     m_Multiplier(multiplier)
 {
 }
 
-DataType MeanAbsoluteDeviationNoiseEstimator::Estimate(const Signal& intensities) const
+DataType MedianAbsoluteDeviationNoiseEstimator::Estimate(Signal& highFreqCoefficients) const
 {
-    constexpr auto locationOfThirdQuartileInNormalDistribution = static_cast<DataType>(.6745);
+    constexpr auto inverseOfThirdQuartileInNormalDistribution = static_cast<DataType>(1.0 / .6745);
```
mg6: Good catch!

src/Spectre.libStatistics.Tests/StatisticsTest.cpp
```diff
@@ -53,6 +53,16 @@ TEST(MeanTest, mean_of_empty_is_zero)
...
 
+TEST(MedianTest, calculates_vector_median)
+{
+    EXPECT_THAT(Median(data), 2.);
+}
+
+TEST(MedianTest, throws_on_empty_vector)
+{
+    EXPECT_THROW(MedianAbsoluteDeviation(empty), spectre::core::exception::EmptyArgumentException);
+}
```
mg6: This should be in line with other functions, esp. `Mean(empty)` above, and return 0 imo.

How do you think @Daroslaw @gmrukwa?

Daroslaw: According to some random experts on statistics:
"[If you have an empty set, then it will have neither a mean, nor a median, nor a mode.](https://www.algebra.com/algebra/homework/Probability-and-statistics/Probability-and-statistics.faq.question.499197.html)"
So I personally see two solutions here; in both functions either throw exception or return `NaN`, with bias on the second solution as it is exactly how Matlab behaves, so I suppose if a software has "Mat" in its name then it has some authority to form opinions about mathematics.

gmrukwa: I'd go for consistency. If we want to go for `Median(empty)` throwing an exception, we should throw exception in each of the functions. Otherwise we will wonder each time, what's the expected behavior. Follow rule of least astonishment :)

On the other hand I know, empty set has no such statistics, however I had simplicity of use in my mind. Otherwise we'll have to do a lot of checks almost at each call point, just to check whether the set is nonempty. I followed DRY, knowing the possible applications.

Sand3r-: Do we vote now or what?

mg6: Discussed during standup meeting; median should return 0 for empty vector, and this behavior should be documented in code.

mg6 (dismissed):
Good job, seems fine.

src/Spectre.libWavelet/WaveletCoefficients.h
```diff
@@ -1,5 +1,6 @@
 /*
 * WaveletCoefficients.h
+* Decouple wavelet coefficients & constants from decomposer
```
gmrukwa: Interesting... :)

Sand3r-: How is it possible that the commit name has suddenly became a license comment ?

gmrukwa: :D

Sand3r-: Guess I'll just remove that and we can proceed with merging :smile: Good catch though, it took my ~7 reads until I realised that this sentence is completely unrelevant.

mg6 (dismissed):

Daroslaw (approved):

PR #26, opened 2018-02-24
Initial implementation of Difference

Daroslaw:
This is an implementation of Matlab's diff function specified here:
https://www.mathworks.com/help/matlab/ref/diff.html

The main difference is support for only 1D data for now, for `mspeaks` needs.

I attempted to perform as least copies and memory reallocations as possible and I'm open for suggestions how to improve this code performance-wise. My first thought is to give up on `minus()` function and perform all the subtractions in an inner loop. Maybe I'll do some benchmarking on it soon.


src/Spectre.libStatistics/Math.h
```diff
@@ -380,4 +380,30 @@ std::vector<DataType> build(size_t size, Generator generator)
...
 }
 
+/// <summary>
+/// Calculates differences between adjacent elements of given container.
+/// </summary>
+/// <param name="data">The container.</param>
+/// <param name="order">Order of difference operation (how many times the difference is executed).</param>
+/// <returns>Vector of calculated differences.</returns>
+ template <class DataType>
+ std::vector<DataType> diff(gsl::span<const DataType> data, uint16_t order = 1)
```
Sand3r-: I'd rename that to Differenciate

gmrukwa: `std::vector<std::remove_const<DataType>::type> diff(gsl::span<DataType> data, uint16_t order = 1)`

src/Spectre.libStatistics/Math.h
```diff
@@ -380,4 +380,30 @@ std::vector<DataType> build(size_t size, Generator generator)
...
+ {
+    size_t length = data.length();
+    std::vector<DataType> result = std::vector<std::remove_const<DataType>::type>(data.begin(), data.end());
```
gmrukwa: `std::vector<std::remove_const<DataType>::type> result(data.begin(), data.end());`

src/Spectre.libStatistics/Math.h
```diff
@@ -380,4 +380,30 @@ std::vector<DataType> build(size_t size, Generator generator)
...
+
+    if (order == 0 || length < 2 || order >= length)
```
gmrukwa: As I agree with first part )`order == 0`, the later are strange to me. Why don't we do usual validation with exception on failure? Or at least something in the documentation, why the behavior is such...

src/Spectre.libStatistics/Math.h
Sand3r-: Aren't you invoking vector copy here?
Think it'd be safer to limit it down to:
```C++
std::vector<std::remove_const<DataType>::type> result(data.begin(), data.end());
```

src/Spectre.libStatistics/Math.h
```diff
@@ -380,4 +380,30 @@ std::vector<DataType> build(size_t size, Generator generator)
     return result;
 }
 
+/// <summary>
+/// Calculates differences between adjacent elements of given container.
```
gmrukwa: Please add information about the source of all assumptions that should be taken into account w.r.t. this function, i.e. page with MATLAB specification of their `diff` function behaviour. It'll be crucial to know, that e.g. output size is input size - 1, while it's not so explicit in tests.

src/Spectre.libStatistics/Math.h
```diff
@@ -380,4 +380,30 @@ std::vector<DataType> build(size_t size, Generator generator)
...
+    std::vector<DataType> result = std::vector<std::remove_const<DataType>::type>(data.begin(), data.end());
+
+    if (order == 0 || length < 2 || order >= length)
+        return result;
+
+    for(auto i = 0u; i < order; ++i)
+    {
+        gsl::span<const DataType> lhs = gsl::as_span(result.data() + 1, result.size() - 1);
+        gsl::span<const DataType> rhs = gsl::as_span(result.data(), result.size() - 1);
+        memcpy(result.data(), std::vector<DataType>(minus(lhs, rhs)).data(), (result.size() - 1) * sizeof(DataType));
```
gmrukwa: We're allocating the memory either way. Why don't we simply do `result = minus(lhs, rhs)`? It would exercise `move` operator, without copying and resizes.

Daroslaw: Nuhuh, such assignment changes the position in memory of `result`. This approach preserves it.

src/Spectre.libStatistics/Math.h
Daroslaw: Alright, I'll do some exceptions.

src/Spectre.libStatistics/Math.h
gmrukwa: But we're allocating the memory in `minus` either way, while constant position of `result` in memory gives us no gain. Am I wrong?

src/Spectre.libStatistics/Math.h
Sand3r-: This could be optimized to
```C++
uint16_t skip = 0 - !(length < 2 || order >= length);
order = order & skip;
```
That way order will be equal to 0 whenever the condition fails, meaning the loop wouldn't execute. 
However, it'd be nice to explain in comment what does that operation do.

src/Spectre.libStatistics.Tests/MathTest.cpp
```diff
@@ -218,4 +218,11 @@ TEST(AbsTest, returns_absolute_value_of_each_element)
     const auto absolutes = abs(gsl::as_span(data));
     EXPECT_THAT(absolutes, ContainerEq(Data{ 1,2,3,4 }));
 }
+
+TEST(DiffTest, calculates_differential)
```
Sand3r-: Would be nice to add tests when an empty vector is passed, or when the length of vector is lower than the requested order.

src/Spectre.libStatistics/Math.h
Sand3r-: I wish we could ```zjeść ciastko i mieć ciastko```, but no exceptions can be thrown without branches :worried:.

src/Spectre.libStatistics/Math.h
Daroslaw: That is why I considered doing it in some other way than using `minus`. It is here mostly just for the same of using already created functionality. I will tinker with `transform`.

src/Spectre.libStatistics/Math.h
Daroslaw: What? Why would you want to change the signature of the function to return type that differs from every single return type of any mathematical function in here? Please elaborate.

src/Spectre.libStatistics/Math.h
Daroslaw: The branching here isn't that painful as it is only a single check, then performing branchless algorithm.

src/Spectre.libStatistics/Math.h
Daroslaw: Done

src/Spectre.libStatistics.Tests/MathTest.cpp
Daroslaw: Done

src/Spectre.libStatistics/Math.h
gmrukwa: Compiler predicts branches preety well :)

src/Spectre.libStatistics/Math.h
```diff
@@ -380,4 +381,43 @@ std::vector<DataType> build(size_t size, Generator generator)
...
+/// </summary>
+/// <param name="data">The container.</param>
+/// <param name="order">Order of difference operation (how many times the difference is executed).</param>
+/// <returns>Vector of calculated differences, of size: length_of_data - order.</returns>
+template <class DataType>
+std::vector<typename std::remove_const<DataType>::type> differentiate(gsl::span<DataType> data, uint16_t order = 1)
+{
+    size_t length = data.length();
+    if (length < 2)
+        throw std::invalid_argument("Data must have length greater than two in order to perform differentiation.");
```
gmrukwa: We have our exception base to differentiate our exceptions of the unhandled ones. We have even `ArgumentOutOfRangeException` or something like this in `spectre::core::exception`. It could be of not too bad use here. (You can specify `data.size()` as bad parameter).

For the future: do not provide user-oriented messages at exception throwing site. Exception should gather the context in such a manner that does not violate DRY and is maintainable. Custom messages at each throw site aren't.

src/Spectre.libStatistics/Math.h
```diff
@@ -380,4 +381,43 @@ std::vector<DataType> build(size_t size, Generator generator)
...
+/// Calculates differences between adjacent elements of given container.
+/// The function is based upon Matlab function diff():
+/// https://www.mathworks.com/help/matlab/ref/diff.html
+/// but it can handle only one dimensional data.
+/// </summary>
+/// <param name="data">The container.</param>
+/// <param name="order">Order of difference operation (how many times the difference is executed).</param>
+/// <returns>Vector of calculated differences, of size: length_of_data - order.</returns>
+template <class DataType>
+std::vector<typename std::remove_const<DataType>::type> differentiate_unsafe(gsl::span<DataType> data, uint16_t order = 1)
```
Sand3r-: We'll need to preapre an issue for revising the naming convention in this header file.

PR #32, opened 2018-02-28
Feature/ga classifier and fitness function

Regis128:
Added:
-new function to OpenCvDataset to get filtered dataset
-GaClassifier class as a new classifier using GA
-GaFitnessFunction - "connector" between data and classifier


Daroslaw (dismissed):
The code looks robust, good job.

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
```diff
@@ -278,4 +278,31 @@ TEST_F(OpenCvDatasetTest, returns_observation_with_valid_entries_through_square_
...
+    const std::vector<Label> result_labels{ 3, 14 };
+    std::vector<bool> include{ 1, 0, 1 };
+    OpenCvDataset filtered = dataset->getFilteredOpenCvDataset(include);
+    gsl::span<const Observation> filtered_data_gsl = filtered.GetData();
+    std::vector<DataType> filtered_data{};
+    for (Observation observation: filtered_data_gsl)
+    {
+        for (DataType dataType: observation)
+        {
+            filtered_data.push_back(dataType);
```
Sand3r-: I have some serious doubts whether the dataType is a good name for something that is not a type.

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
```diff
@@ -278,4 +278,31 @@ TEST_F(OpenCvDatasetTest, returns_observation_with_valid_entries_through_square_
         EXPECT_EQ(data[i], observation[i]);
     }
 }
+
+TEST_F(OpenCvDatasetTest, returns_correct_filtered_data)
+{
+    const std::vector<DataType> result_data{ 0.5f, 0.4f, 0.6f, 2.1f, 1.0f, 0.6f };
+    const std::vector<Label> result_labels{ 3, 14 };
```
Sand3r-: I'd rather move the definitions and declarations of these variable right before their usage, since their current placement obfuscates the mental cache of the reader.

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
```diff
@@ -278,4 +278,31 @@ TEST_F(OpenCvDatasetTest, returns_observation_with_valid_entries_through_square_
...
+    std::vector<bool> include{ 1, 0, 1 };
+    OpenCvDataset filtered = dataset->getFilteredOpenCvDataset(include);
+    gsl::span<const Observation> filtered_data_gsl = filtered.GetData();
+    std::vector<DataType> filtered_data{};
+    for (Observation observation: filtered_data_gsl)
```
Sand3r-: Is there a difference between initializing the filtered data with constructor using begin and end iterators?
```C++
std::vector<DataType> filtered_data(filtered_data_gsl.begin(), filtered_data_gsl.end());
```

src/Spectre.libGaClassifier.Tests/GaClassifierTest.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include "Spectre.libClassifier/Types.h"
+#include "Spectre.libClassifier/Svm.h"
+#include "Spectre.libGaClassifier/GaClassifier.h"
+
+namespace spectre::supervised
+{
+
+TEST(GaClassifierInitialization, initializes)
+{
+    const double TRAINING_SET_SPLIT_RATE = 0.7;
```
Sand3r-: Something to consider:
1. Changing consts to consexprs
2. Since these constants repeat in test class and here, perhaps it'd be nice to extract them to be globals to not repeat the code.

src/Spectre.libGaClassifier/GaClassifier.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+void GaClassifier::Fit(LabeledDataset dataset)
+{
+    const auto& data = asSupported(dataset);
+    RandomSplitter splitter(m_TrainingDatasetSizeRate, m_Seed);
+    auto splittedDataset = splitter.split(data);
+    auto trainingSetSize = splittedDataset.trainingSet.size();
+
+    IndividualFeasibilityConditionsFactory conditionsFactory(splittedDataset.trainingSet.GetSampleMetadata(), splittedDataset.trainingSet.size(), m_MinimalFillup, m_MaximalFillup);
```
Sand3r-: Think you could replace  `splittedDataset.trainingSet.size()` with `trainingSetSize` since it's shorter.

src/Spectre.libGaClassifier/GaClassifier.h
```diff
@@ -0,0 +1,113 @@
...
+#include <vector>
+#include "Spectre.libGenetic/DataTypes.h"
+#include "Spectre.libClassifier/IClassifier.h"
+#include "Spectre.libGenetic/GeneticAlgorithmFactory.h"
+
+namespace spectre::supervised
+{
+    class OpenCvDataset;
+
+    /// <summary>
```
Sand3r-: This blanks space :D

Sand3r- (dismissed):
Nicely done :). I have added a few remarks for you to consider.

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
gmrukwa: This applies always! Good catch!

src/Spectre.libClassifier/OpenCvDataset.cpp
```diff
@@ -121,7 +122,22 @@ gsl::span<const Label> OpenCvDataset::GetSampleMetadata() const
     return m_labels;
 }
 
-size_t OpenCvDataset::size() const
+OpenCvDataset OpenCvDataset::getFilteredOpenCvDataset(std::vector<bool> filterPattern) const
```
gmrukwa: **Major**

Passing by value. This will be ultra inefficient.

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
```diff
@@ -278,4 +278,31 @@ TEST_F(OpenCvDatasetTest, returns_observation_with_valid_entries_through_square_
...
+    {
+        for (DataType dataType: observation)
+        {
+            filtered_data.push_back(dataType);
+        }
+    }
+    gsl::span<const Label> filtered_labels_gsl = filtered.GetSampleMetadata();
+    std::vector<Label> filtered_labels(filtered_labels_gsl.begin(), filtered_labels_gsl.end());
+    for (auto i = 0u; i < result_data.size(); i++)
```
gmrukwa, line 301: **Minor**

Use `Pointwise` matcher. @Sand3r- has already done this, so he will tell you, how to use this with reasonable data, or where to look for this.

src/Spectre.libClassifier/OpenCvDataset.h
```diff
@@ -81,6 +81,13 @@ class OpenCvDataset final : public spectre::core::dataset::IReadOnlyDataset<Obse
     /// <returns></returns>
     gsl::span<const Label> GetSampleMetadata() const override;
 
+    /// <summary>
+    /// Gets part of dataset based on vector of bool with information
+    /// which observation and label to leave
+    /// </summary>
+    /// <returns></returns>
+    OpenCvDataset getFilteredOpenCvDataset(std::vector<bool> filterPattern) const;
```
gmrukwa: **Major**

Do not grow this `OpenCvDataset` more please. Extract this to separate module, or we will end up with another *God Object* as it was with first version of Dataset for C# some time ago. This can be even a standalone function to not to make a lot of things with introducing a class or something.

You can even declare an `operator[](const std::vector<bool>&)`, but do this externally. You need no access to the internals of `OpenCvDataset`. Moreover, you can declare this for any `IDataset` in one place probably (but still returning an `OpenCvDataset`.

src/Spectre.libGaClassifier/GaFitnessFunction.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include "GaFitnessFunction.h"
+#include "Spectre.libException/InconsistentArgumentSizesException.h"
+#include "Spectre.libFunctional/Filter.h"
+#include "Spectre.libClassifier/ConfusionMatrix.h"
+
+namespace spectre::supervised
+{
+
+GaFitnessFunction::GaFitnessFunction(IClassifier& classifier, SplittedOpenCvDataset& data):
```
gmrukwa: **Minor**

I'd say it's a `ClassifierFitnessFunction`.

It doesn't matter whether `FitnessFunction` is used with GA or other type of optimizer (since we will potentially use others!). What matters is that it scores `IClassifier`.

src/Spectre.libGaClassifier/GaFitnessFunction.h
```diff
@@ -0,0 +1,64 @@
...
+/// </summary>
+class GaFitnessFunction : public spectre::algorithm::genetic::FitnessFunction
+{
+public:
+    /// <summary>
+    /// Initializes a new instance of the <see cref="GaFitnessFunction"/> class.
+    /// </summary>
+    /// <param name="classifier">The classifier.</param>
+    /// <param name="data">The splitted data.</param>
+    GaFitnessFunction(IClassifier& classifier, SplittedOpenCvDataset& data);
```
gmrukwa: **CRITICAL**

If we start running evaluations of `GaFitnessFunction` for `IClassifier&`, we are operating on single instance. Imagine, how much it blows up in parallel :)

We already have it called in parallel!

I'd consider consuming the instance here (probably as `std::unique_ptr<IClassifier>`) or just remembering it (as `const IClassifier&`), and then copying it for each evaluation. This should be lightweight, since this classifier before fitting should consist of the parameterization only, not a model itself. @Sand3r-, what do you think about it?

This is something I was wondering about some time ago: to separate parameterization & creation of an instance of `IClassifier` from actual model. We would have no such problem now, as it would just be functional-manner.

gmrukwa (changes requested):
I think that some conceptual changes are required. While code itself is pretty good, we have to keep paralelism in mind, or we will blow up :)

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
Sand3r-: 1. You'll need this [file](https://github.com/spectre-team/native-algorithms/blob/master/src/Spectre.libWavelet.Tests/FloatingPointVectorMatcher.h).
2. [Usage](https://github.com/spectre-team/native-algorithms/blob/master/src/Spectre.libWavelet.Tests/WaveletReconstructorRefTest.cpp#L60)

Sand3r-: Btw. while you are on it, you could possibly extract this floating point matcher header to the common folder and make other projects link to it. That's optional though, since it's out of scope of this review, but it'd be good to make a task for it on trello.

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
Regis128: done

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
Regis128: done

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
Regis128: done

Regis128: but this won't work, so it has to be like that

src/Spectre.libGaClassifier.Tests/GaClassifierTest.cpp
Regis128: done, but we always used just const and I don't see the need for constexpr

src/Spectre.libGaClassifier/GaClassifier.cpp
Regis128: done

src/Spectre.libGaClassifier/GaClassifier.h
Regis128: done, nice catch

src/Spectre.libClassifier/OpenCvDataset.cpp
Regis128: done

src/Spectre.libGaClassifier/GaFitnessFunction.cpp
Regis128: done

src/Spectre.libGaClassifier/GaFitnessFunction.h
Regis128: done

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
Regis128: Don't want to create new dependency just for one test. Maybe in the future when we will create independent module for that

src/Spectre.libClassifier/OpenCvDataset.h
Regis128: done

gmrukwa:
[Merge updates of GA](https://trello.com/c/rHY973Dc/344-merge-updates-of-ga)

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
gmrukwa: @Regis128, then we should add another branch (right now) to extract it. We're spamming such code everywhere. It's better to move this topic to separate branch to not to pollute this review, but if we don't stick to consistent way of testing now, we'll end with bunch of duplicates.

appveyor.yml
```diff
@@ -13,7 +13,7 @@ environment:
 configuration:
   - Debug
   - Release
-version: 3.14.0.{build}
+version: 3.15.0.{build}
```
gmrukwa: Since you've extended the definition of the interface `IClassifier`, this is breaking change now. Please update to `4.0.0`.

src/Spectre.libClassifier/DatasetFilter.h
```diff
@@ -0,0 +1,41 @@
+/*
+* DatasetFilter.h
+* ENTER CLASS DESCRIPTION HERE.
```
gmrukwa: Yeah... Enter the description please :)

src/Spectre.libClassifier/DatasetFilter.h
```diff
@@ -0,0 +1,41 @@
...
+*/
+
+#pragma once
+#include <vector>
+#include "Spectre.libFunctional/Filter.h"
+#include "OpenCvDataset.h"
+
+namespace spectre::supervised
+{
+inline OpenCvDataset getFilteredOpenCvDataset(const OpenCvDataset* dataset, std::vector<bool>& filterPattern)
```
gmrukwa: Not really sure, whether this should ever be `inline`. @Sand3r-?

gmrukwa: You're passing `const OpenCvDataset*`. What if `dataset == nullptr`? Either use `const OpenCvDataset&` or check for `nullptr`. The first would decrease branching.

src/Spectre.libClassifier/DatasetFilter.h
```diff
@@ -0,0 +1,41 @@
...
+{
+    const auto& dataToFilter = dataset->GetData();
+    const auto twoDimensionalFilteredData = core::functional::filter(dataToFilter, filterPattern);
+    std::vector<DataType> oneDimensionalFilteredData;
+    oneDimensionalFilteredData.reserve(twoDimensionalFilteredData.size() * twoDimensionalFilteredData[0].size());
```
gmrukwa: What if everything is filtered out? Better use `(*dataset)[0].size()` instead of `twoDimensionalFilteredData[0].size()`.

src/Spectre.libClassifier/DatasetFilter.h
gmrukwa: Are you modifying `filterPattern`? If not, it should be `const std::vector<bool>&`.

gmrukwa (changes requested):
Major version number must be updated.

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
Regis128: ok, I will do it in a minute

appveyor.yml
Regis128: done

src/Spectre.libClassifier/DatasetFilter.h
Regis128: you're right, done

src/Spectre.libClassifier/DatasetFilter.h
Regis128: done

src/Spectre.libClassifier/DatasetFilter.h
Regis128: hmmm, maybe not, fixed

src/Spectre.libClassifier/DatasetFilter.h
Regis128: nice catch, done

src/Spectre.libClassifier/DatasetFilter.h
Regis128: done

Regis128: ok, without it it fails compilation, and besides i read that inline function should be used when performance is needed

src/Spectre.libClassifier.Tests/OpenCvDatasetTest.cpp
Regis128: created feature/pointwise-test-fix branch for this

gmrukwa: It should be named fixup/pointwise-test, but nevermind :D :D :D

src/Spectre.libClassifier/DatasetFilter.h
gmrukwa: `inline` is used for short functions, as code marked as `inline` is pasted everywhere instead of function call (there are slight exceptions, depending on compiler). For longer functions it's not so simple. But let's leave it now as it is.

gmrukwa (approved):

Daroslaw (approved):

PR #33, opened 2018-03-04
Implementation of algorithm for finding full-width half-height markers for peaks.

Daroslaw:
Suggestions related to performance improvement are warmly welcomed.

This class will be used only for peak finding algorithms and is a part of larger pipeline, so for performance purposes the input data that goes into method of this class is assumed to be verified during earlier stages of the whole pipeline. More information in the methods documentation in header file.

### DIFF
--- a/src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
+++ b/src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
@@ -29,6 +29,16 @@ namespace
 
     class FWHHCalculatorTest : public Test
     {
+    public:
+        FWHHCalculatorTest()
+        {
+            x = { 1., 2., 3., 4., 5., 6., 7., 8., 9. };
+
+            sawtoothY = { 0., 1., 2., 1., 0., 1., 2., 1., 0. };
+            sawtoothValleyIdx = { 0, 4, 8 };
+            sawtoothPeakIdx = { 2, 6 };
+        }
+
     protected:
         FWHHCalculator fwhh;
         Signal x;
@@ -36,17 +46,6 @@ namespace
         Signal sawtoothY;
         Indices sawtoothValleyIdx;
         Indices sawtoothPeakIdx;
-
-        void SetUp() override
-        {
-            fwhh = FWHHCalculator();
-
-            x = { 1., 2., 3., 4., 5., 6., 7., 8., 9. };
-
-            sawtoothY = { 0., 1., 2., 1., 0., 1., 2., 1., 0. };
-            sawtoothValleyIdx = { 0, 4, 8 };
-            sawtoothPeakIdx = { 2, 6 };
-        }
     };
 
     TEST_F(FWHHCalculatorTest, sawtooth_correct_left_fwhh)
--- a/src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
+++ b/src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
@@ -37,6 +37,10 @@ namespace
             sawtoothY = { 0., 1., 2., 1., 0., 1., 2., 1., 0. };
             sawtoothValleyIdx = { 0, 4, 8 };
             sawtoothPeakIdx = { 2, 6 };
+
+            skewedY = { 2., 0., 1., 2., 3., 4., 2., 0., 1. };
+            skewedValleyIdx = { 1, 7 };
+            skewedPeakIdx = { 5 };
         }
 
     protected:
@@ -46,6 +50,10 @@ namespace
         Signal sawtoothY;
         Indices sawtoothValleyIdx;
         Indices sawtoothPeakIdx;
+
+        Signal skewedY;
+        Indices skewedValleyIdx;
+        Indices skewedPeakIdx;
     };
 
     TEST_F(FWHHCalculatorTest, sawtooth_correct_left_fwhh)
@@ -60,5 +68,19 @@ namespace
         Signal result = fwhh.GetRightFWHH(as_span(x), as_span(sawtoothY),
             as_span(sawtoothValleyIdx), as_span(sawtoothPeakIdx));
         EXPECT_THAT(result, ContainerEq(Signal{ 4., 8. }));
+    }
+
+    TEST_F(FWHHCalculatorTest, skewed_correct_left_fwhh)
+    {
+        Signal result = fwhh.GetLeftFWHH(as_span(x), as_span(skewedY),
+            as_span(skewedValleyIdx), as_span(skewedPeakIdx));
+        EXPECT_THAT(result, ContainerEq(Signal{ 4. }));
+    }
+
+    TEST_F(FWHHCalculatorTest, skewed_correct_right_fwhh)
+    {
+        Signal result = fwhh.GetRightFWHH(as_span(x), as_span(skewedY),
+            as_span(skewedValleyIdx), as_span(skewedPeakIdx));
+        EXPECT_THAT(result, ContainerEq(Signal{ 7. }));
     }
 }
--- a/src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
+++ b/src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
@@ -1,0 +1,65 @@
+/*
+* FWWHHCalculatorTest.cpp
+* Test for calculations of full width at half height markers.
+*
+Copyright 2018 Dariusz Kuchta
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <gtest/gtest.h>
+#include "Spectre.libPeakFinder/FWHHCalculator.h"
+#include <gmock/gmock-matchers.h>
+
+namespace
+{
+    using namespace spectre::algorithm::peakfinder;
+    using namespace testing;
+    using namespace gsl;
+
+    class FWHHCalculatorTest : public Test
+    {
+    protected:
+        FWHHCalculator fwhh;
+        Signal x;
+
+        Signal sawtoothY;
+        Indices sawtoothValleyIdx;
+        Indices sawtoothPeakIdx;
+
+        void SetUp() override
+        {
+            fwhh = FWHHCalculator();
+
+            x = { 1., 2., 3., 4., 5., 6., 7., 8., 9. };
+
+            sawtoothY = { 0., 1., 2., 1., 0., 1., 2., 1., 0. };
+            sawtoothValleyIdx = { 0, 4, 8 };
+            sawtoothPeakIdx = { 2, 6 };
+        }
+    };
+
+    TEST_F(FWHHCalculatorTest, sawtooth_correct_left_fwhh)
+    {
+        Signal result = fwhh.GetLeftFWHH(as_span(x), as_span(sawtoothY),
+            as_span(sawtoothValleyIdx), as_span(sawtoothPeakIdx));
+        EXPECT_THAT(result, ContainerEq(Signal{ 2., 6. }));
+    }
+
+    TEST_F(FWHHCalculatorTest, sawtooth_correct_right_fwhh)
+    {
+        Signal result = fwhh.GetRightFWHH(as_span(x), as_span(sawtoothY),
+            as_span(sawtoothValleyIdx), as_span(sawtoothPeakIdx));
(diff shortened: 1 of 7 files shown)


Daroslaw:
[Implement calculation of FWHH markers.](https://trello.com/c/TTSeenUc/402-implement-calculation-of-fwhh-markers)

src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+    protected:
+        FWHHCalculator fwhh;
+        Signal x;
+
+        Signal sawtoothY;
+        Indices sawtoothValleyIdx;
+        Indices sawtoothPeakIdx;
+
+        void SetUp() override
```
Sand3r-: Is it necessary to reassign the data upon each test? Perhaps we could move that to the test constructor.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include "FWHHCalculator.h"
+#include "Spectre.libFunctional/Transform.h"
+
+namespace spectre::algorithm::peakfinder
+{
+    Signal FWHHCalculator::GetLeftFWHH(const SignalView x, const SignalView y, const IndicesView valleys, const IndicesView peaks)
+    {
+        Signal result;
+        result.reserve(peaks.size());
```
Sand3r-: It'd be more efficient to resize data and access by index instead of reserving, because on each push_back the length of vector needs to be increased and check for potential memory out-of-bounds write needs to be performed.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
```diff
@@ -0,0 +1,92 @@
...
+            result.push_back(lfwhhVal);
+        }
+
+        return result;
+    }
+
+    Signal FWHHCalculator::GetRightFWHH(const SignalView x, const SignalView y, const IndicesView valleys, const IndicesView peaks)
+    {
+        Signal result;
+        result.reserve(peaks.size());
```
Sand3r-: Same here ofc.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
```diff
@@ -0,0 +1,92 @@
...
+    inline Index FWHHCalculator::GetClosestNeighbourIndex(const SignalView & xSorted, DataType value)
+    {
+        Index valueIdx = 0;
+
+        for (unsigned i = 0; i < unsigned(xSorted.size()); ++i)
+        {
+            bool isGreater = xSorted[i] < value;
+            unsigned cond = unsigned(!isGreater) - 1;
+            i = i + (cond & xSorted.size());
+            valueIdx = (i - 1) & cond;
```
Sand3r-: It'd be nice to add comment what these 3 lines do, since it doesn't really come obvious from just casually reading that.

Sand3r-: I mean, even though I'm a co-author of part of that, I still don't quite get it :rofl:

src/Spectre.libPeakFinder/FWHHCalculator.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+        for (int i = 0; i < peaks.size(); ++i)
+        {
+            Index peakIdx = peaks[i];
+            Index valleyIdx = valleys[i + 1];
+            DataType peakVal = y[peakIdx];
+
+            SignalView segment = y.subspan(peakIdx, valleyIdx - peakIdx);
+            DataType rfwhhPeakVal = peakVal * 0.5;
+            Signal segmentInversed = spectre::core::functional::transform<DataType>(segment, [](DataType val) { return -val; });
```
Sand3r-: Rather segmentNegated ;)

Sand3r- (changes requested):
Honestly a great read, most of the things come pretty clear, however there are a few changes to be considered.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
```diff
@@ -0,0 +1,92 @@
...
+    inline Index FWHHCalculator::GetClosestNeighbourIndex(const SignalView & xSorted, DataType value)
+    {
+        Index valueIdx = 0;
+
+        for (unsigned i = 0; i < unsigned(xSorted.size()); ++i)
+        {
+            bool isGreater = xSorted[i] < value;
+            unsigned cond = unsigned(!isGreater) - 1;
+            valueIdx = (i - 1) & cond;
+            i = i + (cond & xSorted.size());
```
Sand3r-: Please, write a comment on what do those lines do.

Sand3r- (dismissed):

src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
```diff
@@ -0,0 +1,64 @@
...
+    using namespace gsl;
+
+    class FWHHCalculatorTest : public Test
+    {
+    public:
+        FWHHCalculatorTest()
+        {
+            x = { 1., 2., 3., 4., 5., 6., 7., 8., 9. };
+
+            sawtoothY = { 0., 1., 2., 1., 0., 1., 2., 1., 0. };
```
gmrukwa, line 37: What about additional case with a little bit skewed signal?

src/Spectre.libPeakFinder/FWHHCalculator.h
```diff
@@ -0,0 +1,72 @@
...
+        /// </summary>
+        /// <param name="x">Container view of m/z values.</param>
+        /// <param name="y">Container view of intensity values.</param>
+        /// <param name="valleys">Container view of valley indices.</param>
+        /// <param name="peaks">Container view of peak indices.</param>
+        /// <returns>New container of m/z values for right sides of FWHH markers for all peaks.</returns>
+        Signal GetRightFWHH(const SignalView x, const SignalView y, const IndicesView valleys, const IndicesView peaks);
+    private:
+        inline Index GetClosestNeighbourIndex(const SignalView& xSorted, const DataType value);
+        inline DataType Lerp(DataType x1, DataType x2, DataType y1, DataType y2, DataType x);   //TODO(dkuchta): Maybe in the future move it to libStatistics?
```
gmrukwa: Can we name it more expressively? Or at least add a docstring? It took me a longer while to figure out what's this.

gmrukwa (dismissed):

src/Spectre.libPeakFinder/FWHHCalculator.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include "FWHHCalculator.h"
+#include "Spectre.libFunctional/Transform.h"
+
+namespace spectre::algorithm::peakfinder
+{
+    Signal FWHHCalculator::GetLeftFWHH(const SignalView x, const SignalView y, const IndicesView valleys, const IndicesView peaks)
+    {
+        Signal result;
+        result.resize(peaks.size());
```
gmrukwa: This can be done in single line.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+        for (int i = 0; i < peaks.size(); ++i)
+        {
+            Index peakIdx = peaks[i];
```
gmrukwa: Most of these may be `const`.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
```diff
@@ -0,0 +1,92 @@
...
+            Index valleyIdx = valleys[i];
+            DataType peakVal = y[peakIdx];
+
+            SignalView segment = y.subspan(valleyIdx, peakIdx - valleyIdx);
+            DataType lfwhhPeakVal = peakVal * 0.5;
+            Signal segmentNegated = spectre::core::functional::transform<DataType>(segment, [](DataType val) { return -val; });
+            Index lfwhhNearestIdx = GetClosestNeighbourIndex(segmentNegated, -lfwhhPeakVal) + valleyIdx;
+            DataType lfwhhVal = Lerp(y[lfwhhNearestIdx], y[lfwhhNearestIdx + 1], x[lfwhhNearestIdx], x[lfwhhNearestIdx + 1], lfwhhPeakVal);
+
+            result[i] = lfwhhVal;
```
gmrukwa: What about extracting calculation of single `result[i]` to an `inline` function of something and just running `transform`?

src/Spectre.libPeakFinder/FWHHCalculator.h
```diff
@@ -0,0 +1,72 @@
...
+        /// 5. Indices of both valleys and peaks are assumed to be reflecting real valleys/peaks.
+        /// </summary>
+        /// <param name="x">Container view of m/z values.</param>
+        /// <param name="y">Container view of intensity values.</param>
+        /// <param name="valleys">Container view of valley indices.</param>
+        /// <param name="peaks">Container view of peak indices.</param>
+        /// <returns>New container of m/z values for right sides of FWHH markers for all peaks.</returns>
+        Signal GetRightFWHH(const SignalView x, const SignalView y, const IndicesView valleys, const IndicesView peaks);
+    private:
+        inline Index GetClosestNeighbourIndex(const SignalView& xSorted, const DataType value);
```
gmrukwa, line 69: How the hell are you able to make `inline` function implementation in `.cpp` file instead of this one?

gmrukwa:
[Investigate peak-finding algorithm](https://trello.com/c/UtEisfMo/187-investigate-peak-finding-algorithm)

src/Spectre.libPeakFinder/FWHHCalculator.h
Sand3r-: https://stackoverflow.com/a/3993003
TL;DR Because the function is used only inside this cpp file.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
Sand3r-: Totally agree with extracting. Especially since the code in GetRightFWHH is kind of a repetition of what's inside here.

Sand3r-:
Wherever you use operator[] of gsl::span collection, please keep in mind that on each such access, the attempted index is checked against bounds of the array.

To mitigate this, you can:
1. Save the length to a constant
2. Cast the beginning of a span to pointer of underlying data type
3. Iterate over the array using the length and obtained pointer. 

This avoids unnecessary checks and going through unreasonable amount of branches through the program execution.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
Daroslaw: This is very hard to do without doing lots of branching.

src/Spectre.libPeakFinder.Tests/FWHHCalculatorTest.cpp
Daroslaw: Done

Daroslaw:
Holy moly @Sand3r- is soo right. Just look how long it takes to process `operator[]` of `gsl::span`.

gsljestfajny

gmrukwa:
@Daroslaw, @Sand3r-, do you keep in mind, that these `constexpr` functions aren't function calls?

BTW are you sure, that these checks are also happening for `Release`? I know that MS implementation of STL has also a great number of checks & validators, that are just never used in `Release`, only in `Debug`. That's the reason TF works under VS in `Release` but in `Debug` doesn't.

Also, you shouldn't use pointer&size representation, that's what `gsl::span` is an abstraction of. Use `myspan.begin()` and `myspan.end()` if you want to go that way. Going to pointer&size we can throw away `span` in whole solution.

src/Spectre.libPeakFinder/FWHHCalculator.cpp
gmrukwa: Well, no. You are checking only against the end of the collection. The rest of the code is linear. There is no branching.

EDIT: Now I get the point of separating the code repetition, that would introduce branching. If you even would go that way, you can make this function aggressively inlined or with `bool` as template parameter, so compiler will optimize away all `if`s.

Either way: readability over premature optimization. This one is premature. This code will not be ran soooo often.

gmrukwa (dismissed):
Good job! BTW, I've updated: https://github.com/spectre-team/native-algorithms/pull/33#discussion_r173653031

gmrukwa:
Guys, we're cluttering up the code with unpacking of `gsl::span`. Please wait with such premature optimization until we are able to profile the code.

**readability > premature optimization**

Daroslaw:
Commit [`17651c0`](https://github.com/spectre-team/native-algorithms/commit/17651c095f4a11a57965aad5a6b528f7fa4d1057) nuked.

gmrukwa (approved):

Sand3r-:
I did some profiling, and it looks like there is truly no performance hit. (At least on small chunks of data).
```
Run on (8 X 2594 MHz CPU s)
03/11/18 15:21:49
Benchmark           Time           CPU Iterations
-------------------------------------------------
BM_FWHHgsl        253 ns        248 ns    2639500
BM_FWHHraw        255 ns        254 ns    2639500
```

Sand3r- (approved):

PR #35, opened 2018-03-04
Implement Extrema Finder

Sand3r-:
ExtremaFinder facility allows for finding minima (valleys) and maxima (peaks) of the function of interest.


Sand3r-:
[Implement Extrema Finding](https://trello.com/c/2UAqIJpk/401-implement-extrema-finding)

src/Spectre.libFunctional/Filter.h
```diff
@@ -33,14 +33,14 @@ namespace spectre::core::functional
 /// <param name="collection">The collection.</param>
 /// <param name="indexes">The indexes.</param>
 /// <returns>Vector consisting of elements under specified indexes in the input collection</returns>
-template <typename DataType>
-std::vector<DataType> filter(gsl::span<const DataType> collection, gsl::span<const size_t> indexes)
+template <typename DataType, typename Index>
+std::vector<DataType> filter(gsl::span<const DataType> collection, gsl::span<const Index> indexes)
```
Daroslaw, line 37: Sudden changes in other projects are always welcome

Sand3r-:
Will increment appveyor version upon approval.

Daroslaw (dismissed):
Great job. I wonder if testing the assumption that number of valleys must be greater by one than number of peaks or should it be done when everything is tested as an entire pipeline.

src/Spectre.libFunctional/Filter.h
Sand3r-: Right, I'll make a PR for that next time :smile:

Sand3r-:
Good idea with the peak - valley ratio. Will implement in a moment.

src/Spectre.libFunctional/Filter.h
gmrukwa: There's not lot of it, so let's take this reasonably :P

src/Spectre.libPeakFinder.Tests/ExtremaFinderTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    class ExtremaFinderTest : public ::testing::Test
+    {
+    public:
+        ExtremaFinderTest()
+        {
+            signal = {
+                1.0, 2.0, 2.5, 2.8, 3.5, 2.8, 4.0, 3.0,
+                3.0, 1.0, 1.5, 2.0, 2.5, 2.0, 1.0, 0.0
```
gmrukwa, line 39: Interesting case. Good that it's not the most straightforward one :)

src/Spectre.libPeakFinder/DataTypes.h
```diff
@@ -0,0 +1,33 @@
+/*
```
gmrukwa, line 1: I'm pretty sure I've seen this file in #33.

src/Spectre.libPeakFinder/ExtremaFinder.cpp
```diff
@@ -0,0 +1,100 @@
...
+// Branchless cache-efficient zero filter
+static inline Indices GetNonZeros(const Signal& data)
+{
+    const size_t length = data.size();
+    std::vector<unsigned> indices(length + 1);
+    unsigned zeros = 0;
+
+    for (unsigned i = 0; i < length; i++)
+    {
+        bool dataIsZero = data[i] == 0.0;
```
gmrukwa, line 37: Comparison of `double` by means of `==`.

src/Spectre.libPeakFinder/ExtremaFinder.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    return indices;
+}
+
+// This function finds indices of intensities that combined together create signal with removed flat lines.
+static inline Indices FilterFlatSegmentIndices(const SignalView& signal)
+{
+    Signal derivatives = differentiate(signal);
+    Indices indicesOfNonZeros = GetNonZeros(derivatives);
+    indicesOfNonZeros.push_back((unsigned)(signal.size() - 1));
```
gmrukwa, line 53: Above you've resized the `indices` to have no place for this element for sure. You'd spare reallocation if you managed to keep its reserved size one element more.

gmrukwa (dismissed):
* One of the best, so comprehensive test fixtures I've ever seen.
* Readable, clean code.
* Nice idea with `FindExtrema` for deduplication of code. I wouldn't insist when we have single duplicate in the same place as original, but it definitely widens the horizons to see it done so smoothly.

Really good job, just few minor remarks left for some considerations.

The only thing I am really worried about in long run is the comparison of `double` using `==`.

gmrukwa:
[Investigate peak-finding algorithm](https://trello.com/c/UtEisfMo/187-investigate-peak-finding-algorithm)

src/Spectre.libPeakFinder.Tests/ExtremaFinderTest.cpp
Sand3r-: Thanks! I've used https://plot.ly/create/ to visualise the data during creation.

src/Spectre.libPeakFinder/DataTypes.h
Sand3r-: Yeah, it comes form the same commit. We have also considered moving the DataTypes.h to some common folder, since so many projects make use of it. It'd also make it easier to test the pipeline for the float types, since it'd render just a single file to be changed.

src/Spectre.libPeakFinder/ExtremaFinder.cpp
Sand3r-: That's fine, because in half_float/float/double representation zero is represented by either all bits set to zero, or same, but the sign bit being set to one. Both cases are handled nicely though.

src/Spectre.libPeakFinder/ExtremaFinder.cpp
Sand3r-: So the call
```C++
indices.reserve(length - zeros);
```
Would shrink the vector length but keep the allocated spot for the new push_back?

src/Spectre.libPeakFinder/DataTypes.h
gmrukwa: I think this could be resolved in near future to go for extraction of this file. Not necessarily in this PR.

src/Spectre.libPeakFinder/ExtremaFinder.cpp
gmrukwa: Yeah, but what about potential problems with precision during numeric operations?

src/Spectre.libPeakFinder/ExtremaFinder.cpp
gmrukwa: Actually `indices.reserve(length - zeros + 1);` would have enough place allocated for this additional element, but this still results in a reallocation after you shrinked. I'm not sure whether we can achieve it reasonably, I was just throwing an idea ;) I am not sure, whether `reserve` would do the work without `resize`.

EDIT: **There is no reallocation going down with size**: http://en.cppreference.com/w/cpp/container/vector/resize

Topic closed.

src/Spectre.libPeakFinder/ExtremaFinder.cpp
Sand3r-: If something would happen to not be equal to zero, then it'd either be slightly positive or negative (meaning the function is either growing or decaying), which is completely fine, and plays nicely with the rest of the algorithm

Sand3r-: The only operation that is applied on data in question is subtraction (using differentiate), so as long as the certain peaks have equal intensities there are no rounding errors.

Sand3r-: It'll for sure not make the sign of the result change, and this is the most critical part pretty much. Actually, perhaps a thing worth considering would be to only store sign bits, instead of whole magnitudes.

Daroslaw (approved):

PR #36, opened 2018-03-11
Standarize usage of DataTypes among PeakFinding, GMM and Wavelet projects

Sand3r-:
I have moved the commonly used data types to, ehm, Common/DataTypes.h.
Those projects which used to use their copies now refer to this file. If some projects introduced their own, they just include the Common/DataTypes.h to their own local DataTypes.h and introduce their project-specific types.

Also the following types have been renamed:
`Signal -> Data`
`SignalView -> DataView`

TODO: Need to consult with @Regis128 if it'd be safe to apply the file for his projects.


src/Spectre.libGaussianMixtureModelling.Tests/ExpectationMaximizationTest.cpp
```diff
@@ -34,8 +34,8 @@ class ExpectationMaximizationTest : public ::testing::Test
 {
 protected:
 
-    std::vector<double> mzs;
-    std::vector<double> intensities;
+    std::vector<DataType> mzs;
```
Daroslaw: Whoa, you already defined `std::vector<DataType>` as `Data` so why aren't you using it? The same goes to all repetitions across the changed files.

Daroslaw (changes requested):

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
```diff
@@ -21,7 +21,7 @@ limitations under the License.
 #pragma once
 #include <random>
 #include "Spectre.libException/NullPointerException.h"
-#include "Spectre.libGaussianMixtureModelling/DataType.h"
+#include "Common/DataTypes.h"
```
gmrukwa: Please make sure it is included into our NuGet package. I still wonder whether we shouldn't just have such file inside `Spectre.libDataset` or such.

src/Spectre.libGaussianMixtureModelling/GaussianMixtureModel.h
```diff
@@ -20,6 +20,7 @@
 #pragma once
 #include <vector>
 #include <gsl.h>
+#include "../Common/DataTypes.h"
```
gmrukwa: Should be w.r.t. root of solution.

src/Spectre.libGaussianMixtureModelling/RandomInitializationRef.h
```diff
@@ -34,7 +34,7 @@ limitations under the License.
 #include <vector>
 #include <random>
 #include "Spectre.libGaussianMixtureModelling/GaussianMixtureModel.h"
-#include "Spectre.libGaussianMixtureModelling/DataType.h"
+#include "Common/DataTypes.h"
 
 typedef std::mt19937_64 RandomNumberGenerator;
```
gmrukwa: This should also go to this type definition file.

gmrukwa (commented):
I am still not so sure about replacing every occurence of `Signal` and other entities by a meaningless `Data`. Don't we loose context?

Sand3r-:
> I am still not so sure about replacing every occurence of Signal and other entities by a meaningless Data. Don't we loose context?

Well, I was afraid of that too, but when I was replacing the type names, I've noticed that pretty much all variable names are good and indicate what are they designed to do. Most old `Signal`s are called `signal`.

src/Spectre.libGaussianMixtureModelling/GaussianMixtureModel.h
Sand3r-: I tried doing that, but for some reason the IDE said that it cannot find the path. Now I've found it to compile just fine.

Daroslaw (dismissed):

Daroslaw (approved):

src/Spectre.libGaussianMixtureModelling/GaussianMixtureModel.h
gmrukwa: Then, probably property sheet is not added so you are missing `$(SolutionDir)` in `Addtional Include Directories`. Either way, due to NuGet packaging, it must be w.r.t. root of solution.

src/Spectre.libGaussianMixtureModelling/DataTypes.h
```diff
@@ -16,6 +16,11 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 */
-
 #pragma once
-typedef double DataType;
+#include <random>
+#include "Common\DataTypes.h"
+
+namespace spectre::unsupervised::gmm
+{
+using RandomNumberGenerator = std::mt19937_64;
```
gmrukwa, line 25: I thought about it in `Common\DataTypes.h`. It will be useful also for `libGenetic`.

Sand3r-:
[Make DataTypes header reusable accross projects](https://trello.com/c/zoQhtgYH/409-make-datatypes-header-reusable-accross-projects)

gmrukwa (approved):

src/Spectre.libGaussianMixtureModelling/DataTypes.h
Sand3r-: Yeah I initially thought the same, but currently libGenetic and libClassifier aren't using those types just yet. If we happen to include them, we can certainly move this part.

PR #39, opened 2018-03-13
Splitting peaks finder 

Sand3r-:
The feature introduces a module for finding clear (aka splitting) peaks among the provided list of peaks, using user-supplied parameters and pre-calculated peak qualities.

To shed a bit more light on how does it work:
1. Using FindOffset, the function determines a new interval in which the clear peaks will be searched.
2. All peaks contained in interval are then filtered out if their quality or height aren't satisfactory and the highest of those who have remained is chosen as a clear peak of the interval. Index of that "best" peak is then returned. If there were no peaks satisfying conditions found, function returns a special index indicating that.
3. Go to step 1 and repeat until reaching the end of peak array.

Btw. sorry for FindClearPeaks being this long. Believe it or not, but I've spent ~3-4h on trying to make it smaller, and that's the furthest I could get it to.

### DIFF
--- a/src/Spectre.libGaussianMixtureModelling/ClearPeakFinder.cpp
+++ b/src/Spectre.libGaussianMixtureModelling/ClearPeakFinder.cpp
@@ -18,9 +18,11 @@ limitations under the License.
 */
 #include <stdexcept>
 #include "ClearPeakFinder.h"
+#include "Spectre.libException\EmptyArgumentException.h"
 #include "Spectre.libFunctional\Filter.h"
 
 using namespace spectre::core::functional;
+using namespace spectre::core::exception;
 namespace spectre::unsupervised::gmm
 {
 // Find beginning/end of interval where the clear peaks are searched for.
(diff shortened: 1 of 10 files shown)


src/Spectre.libGaussianMixtureModelling/ClearPeakFinder.cpp
```diff
@@ -0,0 +1,102 @@
...
+            maxIntensity = intensities[i];
+        }
+    }
+    return bestCandidate;
+}
+
+Indices ClearPeakFinder::FindClearPeaks(const Peaks& peaks, const DataView& peakQualities,
+    const GmmOptions& options, DataType resolutionCoeff) const
+{
+    if (peaks.intensities.empty()) throw std::invalid_argument("Empty peak list");
```
Daroslaw: Don't we have any custom Spectre-specific exceptions that would suit those cases?
Something like [this](https://github.com/spectre-team/native-algorithms/blob/40b1d0ace12967d94c25c9e40a2a45ff78b41da0/src/Spectre.libException/EmptyArgumentException.h#L28) for instance.

src/Spectre.libGaussianMixtureModelling/ClearPeakFinder.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (peakQualities.empty()) throw std::invalid_argument("Empty peak quality list");
```
Daroslaw: This function jollily assumes, that length of peak container is equal to length of quality container. Please consider adding a check for that with exception throw.

Daroslaw (changes requested):

Daroslaw (dismissed):

Daroslaw (dismissed):

Regis128 (dismissed):

Daroslaw (approved):

PR #40, opened 2018-03-31
Splitter Segment Extractor

Sand3r-:
Splitter-Segment is a part of a signal that surrounds a desired splitting peak (aka clear peak).
The width of the splitter-segment is determined by some artificial constants.
In order to prevent a splitter-segment to have unnaturally steep start and end, so called margins are added (marked as red).
image
(Source: Supplementary material to *Polanski, Andrzej, et al. "Signal partitioning algorithm for highly efficient Gaussian mixture modeling in mass spectrometry." PloS one 10.7 (2015): e0134256*)

These margins are just halves of normal distribution with appropriately selected variances, and heights set to the first and last intensity of the splitter-segment.


src/Spectre.libGaussianMixtureModelling/SplitterSegmentExtractor.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    return margin;
+}
+
+static inline Data GetIntensitiesOfLeftMargin(SpectrumView spectrum, DataView margin,
+    Index intervalBegin, Index intervalEnd, DataType resolutionCoeff)
+{
+    const DataView& mzs = spectrum.mzs;
+
+    const DataType heightScaleFactor = sqrt(2.0 * M_PI) * 2.0 * resolutionCoeff *
```
Daroslaw: Constant `sqrt(2.0 * M_PI) * 2.0` as `constexpr` maybe?

src/Spectre.libGaussianMixtureModelling/SplitterSegmentExtractor.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    return marginIntensities;
+}
+
+static inline Data GetIntensitiesOfRightMargin(SpectrumView spectrum, DataView margin,
+    Index intervalBegin, Index intervalEnd, DataType resolutionCoeff)
+{
+    const DataView& mzs = spectrum.mzs;
+
+    const DataType heightScaleFactor = sqrt(2.0 * M_PI) * 2.0 * resolutionCoeff *
```
Daroslaw: Same as above.

Daroslaw (dismissed):
Good job! I do believe that you already considered unifying all those mirroring left/right functions and gave up for a good reason, so I approve this code.

src/Spectre.libGaussianMixtureModelling/SplitterSegmentExtractor.cpp
Sand3r-: Any good naming ideas? I've already spent too much trying to come up with one that would make sense. Good compiler (not MSVC) should automatically detect this as constant and compute it in compile time.

Sand3r-:
I gave up on merging the left/right counterparts, because it'd only yield the code less readable, since the discrepancies  between the corresponding functions are much bigger than it used to be in a case of extrema finder.

Thanks for the review btw!

src/Spectre.libGaussianMixtureModelling/SplitterSegmentExtractor.cpp
Daroslaw: `heightScaleConstant`?

Sand3r-: Not all that informative isn't it? :smile:

Regis128 (dismissed):
Good job, seems ok to me

Daroslaw (approved):

PR #41, opened 2018-04-11
Feature/dataset factory

Regis128:
Added DatasetFactory class - for creating OpenCvDataset from external sources, for now files
Also added FileOpenException


src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+        datasetFactory() {}
+protected:
+    const std::vector<DataType> data_compare{ 1.1f, 4.5f, 6.7f, 5.0f, 9.5f, 3.9f, 7.1f, 5, 3.2f, 10, 24.5f, 5.756f,
+        5.09f, 9.6f, 8.23452f, 2, 5.53f, 3.76f, 6.11234f, 6.9f, 7.0f, };
+    const std::vector<Label> labels_compare{ 0, 0, 1, 1, 0, 0, 1 };
+    const std::string filename = "test_file.txt";
+    std::fstream file;
+    DatasetFactory datasetFactory;
+
+    void SetUp()
```
Daroslaw: Can this be performed only once?

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+    {
+        file.open(filename, std::fstream::out);
+        file << "test file" << std::endl;
```
Daroslaw: Can this be done in some more elegant way? Like taken from the data vector and appended in a loop. You could have spared yourself quite a lot time and LoC by that.

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+TEST_F(DatasetFactoryTest, compare_dataset_from_file_and_from_data)
+{
+    OpenCvDataset datasetCompare(data_compare, labels_compare);
+    OpenCvDataset datasetFromFile = datasetFactory.create(filename);
+    for (auto i = 0u; i < datasetCompare.size(); i++)
+    {
+        for (auto j = 0; j < datasetCompare[i].size(); j++)
+        {
+            EXPECT_EQ(datasetCompare[i][j], datasetFromFile[i][j]);
```
Daroslaw: `EXPECT_FLOAT_EQ` maybe?

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,69 @@
...
+OpenCvDataset DatasetFactory::create(std::string filename) const
+{
+    std::fstream file;
+    file.open(filename, std::fstream::in);
+    if (!file.is_open())
+    {
+        throw exception::FileOpenException(filename);
+    }
+    std::vector<DataType> m_Data = {};
+    std::vector<Label> m_labels = {};
```
Daroslaw: These are not members, so by convention they should not be prefixed with `m_`

src/Spectre.libClassifier/DatasetFactory.h
```diff
@@ -0,0 +1,41 @@
...
+public:
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DatasetFactory"/> class.
+    /// </summary>
+    DatasetFactory();
+    /// <summary>
+    /// Creates new OpenCvDataset from file.
+    /// </summary>
+    /// <param name="filename">The filename.</param>
+    OpenCvDataset create(std::string filename) const;
```
Daroslaw: Maybe this could be a static function?

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace spectre::supervised
+{
+const int ColumnMatrixWidth = 1;
+
+DatasetFactory::DatasetFactory()
+{
+}
+
+OpenCvDataset DatasetFactory::create(std::string filename) const
```
Daroslaw: I'm kind of worried that we only check for existence of a text file, but not its validity.

Daroslaw (changes requested):
Good job!
My main issue is related to lack of format validation. So I will accept this commit once validation is provided or if you persuade me that this is a bad idea :)

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
Regis128: good call, moved it to test

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: done

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
Regis128: yeah, it will be more accurate, done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: yes, you're right, I assume we will have resources in correct format. The validity I think should be somewhere else

src/Spectre.libClassifier/DatasetFactory.h
Regis128: It could but it is not necessary I think

src/Spectre.libClassifier/DatasetFactory.cpp
Daroslaw: Let's keep this matter open for discussion with other reviewers.

src/Spectre.libClassifier/DatasetFactory.h
Daroslaw: As the class itself is stateless I think it's worth considering. We will se what others say

gmrukwa: @Daroslaw, you are absolutely right. It is crucial to have this `static`! Otherwise, how would we read the first dataset not having any? 😆

gmrukwa: @Regis128, you almost never want to pass `std::string`, rather than `const std::string&`.

Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
+/*
+* DatasetFactory.cpp
+* Creates OpenCvDataset from external sources.
```
Sand3r-: External sources sound kinda vague, could you please note that it creates a dataset from file?

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+*/
+
+#include <fstream>
+#include <sstream>
+#include "DatasetFactory.h"
+#include "FileOpenException.h"
+
+namespace spectre::supervised
+{
+const int ColumnMatrixWidth = 1;
```
Sand3r-: Minor:
One could use constexpr unsigned for something such as width that can be non-negative. Also, since it is a constant, according to convention it should be composed of uppercase letters with underscores.

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+    std::fstream file;
+    file.open(filename, std::fstream::in);
+    if (!file.is_open())
+    {
+        throw exception::FileOpenException(filename);
+    }
+    std::vector<DataType> data = {};
+    std::vector<Label> labels = {};
+    DataType num;
```
Sand3r-: It is unclear what does this variable represent number of.

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+    Label label;
+    std::string line;
+    {
+        getline(file, line);
+        getline(file, line);
+    }
```
Sand3r-: It'd be nice to at least get some comment on why does the fetching of first 2 lines is required.

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+    while (!file.eof())
+    {
+        //get Label
+        getline(file, line);
+        if (line != "") {
```
Sand3r-: Can be replaced with `line.empty()` for efficiency.

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+            std::istringstream ssLabels(line);
+            while (ssLabels >> label) {}
+            labels.push_back(label);
```
Sand3r-: Not sure what do the 2 lines above do.
Does the 
`while (ssLabels >> label) {}`
save the last "word" to `label` which is then added to the `labels` list?

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+            //get DataTypes
```
Sand3r-: The comment is a bit misleading. After discovering what it does I understood what you've meant, but the code doesn't really get any DataTypes here, instead it just gathers the data.

src/Spectre.libClassifier/DatasetFactory.cpp
Sand3r-: If so, could you please comment on that? It doesn't come obvious just from reading that bit, at least for me.

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+            getline(file, line);
+            std::istringstream ssData(line);
+            while (ssData >> num)
+            {
+                data.push_back(num);
+            }
+        }
+    }
+    OpenCvDataset openCvDataset(data, labels);
+    return openCvDataset;
```
Sand3r-: You can just return OpenCvDataset(data, labels); at this point, for shorter syntax.

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+namespace spectre::supervised
+{
+const int ColumnMatrixWidth = 1;
+
+DatasetFactory::DatasetFactory()
+{
+}
+
+OpenCvDataset DatasetFactory::create(const std::string& filename)
+{
```
Sand3r-: Is the validity of the file checked somewhere earlier (in higher layers of abstraction) in the code? If so, please mention in the dosctrings of the method that the method performs no validation.

Sand3r-: Also another note on the method. To improve the readability of the code, you can extract the code gathering the labels and the data to respective functions (they can either be made private, or static inlines).

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+TEST_F(DatasetFactoryTest, compare_dataset_from_file_and_from_data)
+{
+    file.open(filename, std::fstream::out);
+    for (std::string str: dataForFile)
+    {
+        file << str << std::endl;
+    }
+    file.close();
```
Sand3r-, line 65: It'd be nice to delete the file once it is no longer in use.

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+class DatasetFactoryTest : public ::testing::Test
+{
+public:
+    DatasetFactoryTest() :
+        datasetFactory() {}
+protected:
+    const std::vector<DataType> data_compare{ 1.1f, 4.5f, 6.7f, 5.0f, 9.5f, 3.9f, 7.1f, 5, 3.2f, 10, 24.5f, 5.756f,
+        5.09f, 9.6f, 8.23452f, 2, 5.53f, 3.76f, 6.11234f, 6.9f, 7.0f, };
+    const std::vector<Label> labels_compare{ 0, 0, 1, 1, 0, 0, 1 };
```
Sand3r-: Please keep the consistency in naming of the variables (sometimes you use lowercase_underscore, other times a camelCase). The Spectre convention for C++ is to use camelCase for variable names, if that's what you want to stick to.

Daroslaw (changes requested):
I have got no objections apart from the unresolved file validation case. The discussion is still open.

Sand3r- (commented):
Mostly well done, however there are some minor issues to tackle on (mainly regarding the improvements of readability of the code).

gmrukwa:
Regarding file validation: I think that at least throwing an exception is a **must**.

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: I wanted to write more generally for future, but I will change it then, so ok, changed

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: for now case closed as discussed, I added comment explaining shortly

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
Regis128: I disagree, maybe it will be good and practical idea later, but for now I wouldn't be so hasty. We would likely want to run multiple runs on the same file.

src/Spectre.libClassifier.Tests/DatasetFactoryTest.cpp
Regis128: good point, done

src/Spectre.libClassifier/DatasetFactory.cpp
Regis128: I did it only for label, for data it would slow it a little

src/Spectre.libClassifier/DatasetFactory.cpp
```diff
@@ -0,0 +1,80 @@
...
+        //get Label
+        getline(file, line);
+        if (!line.empty()) {
+            //in text files in lines with coordinates and label values, we skip coordinates
+            //and only get the last number (label)
+            labels.push_back(readLabel(line));
+
+            //we push every number in lines in text file containing data to "data" variable
+            getline(file, line);
+            std::istringstream ssData(line);
```
Sand3r-, line 63: Consistency: It's nice to see that you've made a readLabel method that decreased the line-length of this method, but I don't understand why haven't you done so for the data :smile:

Sand3r- (approved):
Just a minor remark, take it into consideration at your will.
Well done with the comments.

Daroslaw (approved):

PR #42, opened 2018-04-11
Creates ClassifierFactory for easier classifier creation

Regis128:
Added classifierFactory class for easier creation of classifiers


src/Spectre.libGaClassifier/ClassifierFactory.h
```diff
@@ -0,0 +1,58 @@
...
+#include "Spectre.libClassifier/Svm.h"
+#include "Spectre.libGenetic/DataTypes.h"
+#include "GaClassifier.h"
+
+namespace spectre::supervised
+{
+/// <summary>
+/// The factory for creating classifiers
+/// </summary>
+class ClassifierFactory
```
Daroslaw, line 30: Static class maybe? This is yet another class that does not store any state, so I think this will be a suitable approach.

src/Spectre.libGaClassifier/ClassifierFactory.cpp
```diff
@@ -0,0 +1,58 @@
...
+    double bitSwapRate,
+    double preservationRate,
+    unsigned generationsNumber,
+    unsigned populationSize,
+    unsigned initialFillup,
+    spectre::algorithm::genetic::Seed seed,
+    size_t minimalFillup,
+    size_t maximalFillup,
+    unsigned int iterationsLimit,
+    double tolerance) const
```
Daroslaw, line 46: That is a lot of parameters. Have you considered putting at least some of them into a struct like `GaClassifierOptions` or something similar?

src/Spectre.libGaClassifier/ClassifierFactory.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    std::unique_ptr<IClassifier> classifier;
+    if (name == "svm")
```
Daroslaw, line 49: Sooo... in every other case an empty pointer to `IClassifier` is passed to the `GaClassifier` constructor?

Daroslaw (changes requested):

PR #45, opened 2018-05-04
Update GSL version (2017-06)

mg6:
**Summary**
Changes included in this merge request update the underlying GSL library.

**Problem**
Currently, the projects use a nupkg version of Microsoft GSL library, which is outdated (2016), consequently preventing cross-platform builds.

**Solution**
Ubuntu/Debian repositories contain a more current version of the library, labeled as version 0~git20170602.1f87ef7-2. Those sources are committed into the repository and enabled in MSVC projects. Necessary refactoring is done to enable a successful build.

**Note to reviewers**
- Please do not review changes to `dep/gsl/` directory, as they contain only extracted files from the GSL project and comprise the majority of this pull request’s content.
- As the pull request affects the whole project, I am adding more reviewers and requesting authors of specific components to pay special attention to changes impacting their code.

Thanks!

PR #46, opened 2018-05-04
Provide CMake support under Linux

mg6:
**Summary**
This merge request adds CMake configuration to the project, enabling Linux builds.

**Prerequisites**
Requires #45 to be merged and source branch to be rebased on top of this merge.

**Scope**
To enable GCC/Clang builds and refactor existing code while preserving MSVC compatibility.

The possibility of *replacing* existing MSVC solution in favor of CMake was not addressed in this pull request *at all*, as the current scope turned out broad already. It is a future goal to *regenerate* MSVC solution from CMake configuration on Windows when the CMake configuration proves sufficient.

**Addressed problems**
- Providing Linux-oriented CMake configuration.
- Fixing of non-portable include paths.
- Removing extra qualification in header files.
- Adding missing header imports.
- Updating unit tests not to be platform-dependent (RNG).
- Refactoring of Individual class to actually inherit std::vector<bool> to overcome problems with overloaded operators.
- Resolving linker errors related to internal (project/project) and external (OpenMP) dependencies.

CMake project configuration lists all project files instead of globbing them. This is the preferred solution because of discoverability problems connected with glob usage; think *explicit is better than implicit*. See https://stackoverflow.com/a/1060061 for further discussion.

**Environment & dependencies**
Project was confirmed to build under GCC ≥ 7.2 and Clang ≥ 5.0 compilers. Changes were tested under up-to-date Debian unstable setup, as well as with Debian testing Docker image.

In order to build the project, the following minimal set of packages needs to be installed (current versions of Debian unstable packages in brackets):
- build-essential — gcc/g++/make tooling [12.5], with default GCC 7 compiler [7.3.0-17]
- g++-8 — required for GCC 8 builds [8.1.0]
- clang-5.0/clang-6.0/clang-7 — required for Clang builds [7~svn323616-1]
- cmake — CMake toolset [3.11.1]
- cmake-extras — additional CMake finders [1.3+17.04.20170310-1]; bundles GMock finder
- libgmock-dev — GMock/GTest library [1.8.0-10]
- libmsgsl-dev — Microsoft GSL header-only library [0~git20170602.1f87ef7-2]
- libomp-dev — Clang’s OpenMP implementation [6.0-2]
- libopencv-dev — OpenCV library [3.2.0+dfsg-4+b4]

**Testing**
One can test the configuration by issuing:
```sh
# make sure deps are installed
apt-get -qq update && apt-get -qq install build-essential [ clang-7 ] [ g++-8 ] \
    cmake cmake-extras libgmock-dev libmsgsl-dev libomp-dev libopencv-dev

# prepare build directory
mkdir -p build && cd build

# select Clang compiler
export CC="/usr/bin/clang-7"     # or: clang-5.0, clang-6.0
export CXX="/usr/bin/clang++-7"  # or: clang++-5.0, clang++-6.0

# or: select GCC compiler
export CC="/usr/bin/gcc-7"       # or: gcc-8
export CXX="/usr/bin/g++-7"      # or: g++-8

# generate project files, build & run unit tests
cmake ..
cmake --build . -- -j$(nproc)
ctest --output-on-failure -j$(nproc)

# or: using make only; ARGS go to ctest
cmake ..
make -j$(nproc) all test ARGS="--output-on-failure -j$(nproc)"
```

**Note to reviewers**
- Make sure the prerequisite described above is already satisfied.
- Again, as the pull request affects the whole project, I am adding more reviewers and requesting authors of specific components to pay special attention to changes impacting their code. Please don’t go through all the code unless you have superhuman powers.
- Please prefer going through the commits one-by-one instead of jumping into the cumulative diff, as it is huge; you have been warned.

Thanks!

### DIFF
--- a/src/Spectre.libGenetic/Individual.cpp
+++ b/src/Spectre.libGenetic/Individual.cpp
@@ -54,28 +54,4 @@ Individual& Individual::operator=(const Individual &other)
     }
 }
 
-std::vector<bool>::reference Individual::operator[](size_t index)
-{
-    if (index < size())
-    {
-        return (*this)[index];
-    }
-    else
-    {
-        throw OutOfRangeException(index, size());
-    }
-}
-
-std::vector<bool>::const_reference Individual::operator[](size_t index) const
-{
-    if (index < size())
-    {
-        return (*this)[index];
-    }
-    else
-    {
-        throw OutOfRangeException(index, size());
-    }
-}
-
 }
--- a/src/Spectre.libGenetic/Individual.cpp
+++ b/src/Spectre.libGenetic/Individual.cpp
@@ -31,11 +31,6 @@ namespace spectre::algorithm::genetic
 Individual::Individual(std::vector<bool> &&binaryData):
     std::vector<bool>(binaryData) { }
 
-const std::vector<bool>& Individual::getData() const
-{
-    return *this;
-}
-
 bool Individual::operator==(const Individual &other) const
 {
     return std::equal(begin(), end(), other.begin(), other.end());
--- a/src/Spectre.libGenetic/Individual.cpp
+++ b/src/Spectre.libGenetic/Individual.cpp
@@ -29,77 +29,28 @@ using namespace spectre::core::exception;
 namespace spectre::algorithm::genetic
 {
 Individual::Individual(std::vector<bool> &&binaryData):
-    m_BinaryData(binaryData) { }
+    std::vector<bool>(binaryData) { }
 
 const std::vector<bool>& Individual::getData() const
 {
-    return m_BinaryData;
-}
-
-std::vector<bool>::reference Individual::operator[](size_t index)
-{
-    if (index < m_BinaryData.size())
-    {
-        return m_BinaryData[index];
-    }
-    else
-    {
-        throw OutOfRangeException(index, m_BinaryData.size());
-    }
-}
-
-std::vector<bool>::const_reference Individual::operator[](size_t index) const
-{
-    if (index < m_BinaryData.size())
-    {
-        return m_BinaryData[index];
-    }
-    else
-    {
-        throw OutOfRangeException(index, m_BinaryData.size());
-    }
-}
-
-std::vector<bool>::iterator Individual::begin()
-{
-    return m_BinaryData.begin();
-}
-
-std::vector<bool>::iterator Individual::end()
-{
-    return m_BinaryData.end();
-}
-
-std::vector<bool>::const_iterator Individual::begin() const
-{
-    return m_BinaryData.begin();
-}
-
-std::vector<bool>::const_iterator Individual::end() const
-{
-    return m_BinaryData.end();
-}
-
-size_t Individual::size() const
-{
-    return m_BinaryData.size();
+    return *this;
 }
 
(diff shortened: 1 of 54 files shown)


gmrukwa:
[Enable Linux build](https://trello.com/c/Vyp9oQ0K/374-enable-linux-build)

src/Spectre.libGenetic/Individual.h
```diff
@@ -26,7 +26,7 @@ namespace spectre::algorithm::genetic
 /// <summary>
 /// Binary representation of an individual chromosome in genetic algorithm.
 /// </summary>
-class Individual
+class Individual : public std::vector<bool>
```
Daroslaw: whoa

src/Spectre.libGenetic/Individual.cpp
```diff
@@ -29,77 +29,28 @@ using namespace spectre::core::exception;
...
-    return m_BinaryData.end();
-}
-
-size_t Individual::size() const
-{
-    return m_BinaryData.size();
+    return *this;
 }
 
 bool Individual::operator==(const Individual &other) const
```
Daroslaw: The [default operator of std::vector](https://stackoverflow.com/a/16422594) won't make a cut?

src/Spectre.libGenetic/Individual.cpp
```diff
@@ -54,28 +54,4 @@ Individual& Individual::operator=(const Individual &other)
     }
 }
 
-std::vector<bool>::reference Individual::operator[](size_t index)
-{
-    if (index < size())
-    {
-        return (*this)[index];
```
Daroslaw: The infinite recursion problem could be solved [in this way](https://stackoverflow.com/a/34466458).

Daroslaw (dismissed):
That is huge, great job mate!

Sand3r- (dismissed):
All of the rework that covered files introduced by me is either done fine, or there are no breaking changes. Good Job @mg6 !

src/Spectre.libGenetic/Individual.cpp
mg6: Interestingly, I could get it to compile with custom `operator==` removed. But additionally removing `operator!=` causes a build failure in GTest.

Let’s leave both as is for now.

src/Spectre.libGenetic/Individual.cpp
mg6: The problem is, there is nothing to override, as `std::vector`’s [compare operators](http://en.cppreference.com/w/cpp/container/vector/operator_cmp) are *non-virtual*. You could only (1) *shadow* them or (2) define the operators outside of class but this applies to 2-argument ops.

`Individual` class was refactored to inherit `std::vector<bool>` to obviate the need for extra operator definitions.

mg6:
@Sand3r- @Daroslaw Addressed the issue with failing crossover operator test on Appveyor, please review again. Thanks!

Sand3r- (approved):

Daroslaw (approved):

gmrukwa:
[Prepare cmake build for native-algorithms](https://trello.com/c/6mFiRHyI/343-prepare-cmake-build-for-native-algorithms)

PR #47, opened 2018-05-16
Setup Travis CI

mg6:
**Summary**
Enables GCC/Clang builds under Travis CI; configuration made for Ubuntu Trusty (14.04).

**Build matrix**
As GCC 7 and Clang 5.0 are the lowest versions of those compilers with wide C++17 support, such build configurations comprise Travis’ build matrix. Care was taken to enforce C++17 standard when running CMake.

**Dependencies**
Unfortunately, Travis CI does not consider running Ubuntu 16.04/18.04 builds production-ready, so CI configuration targets Ubuntu 14.04 at this point. What follows is that several dependencies are heavily outdated under Trusty, like OpenCV and Clang.

_Compilers_
Ubuntu and LLVM projects provide third-party Personal Package Archives (PPAs) that provide new versions of GCC/Clang/libstdc++ packets for Trusty (Ubuntu & LLVM toolchain PPAs); those were used to enable C++17 compilation.

_OpenMP_
It was not possible to find a backport of Clang’s OpenMP package for 14.04 so its predecessor, older Intel’s implementation is used, in form of `libiomp-dev` package.

_GSL_
There exists no pre-packaged version of Microsoft’s GSL for Trusty, so the version from dep/ directory is used under CI, as well as under Windows.

_GTest_
As Google Test project absorbed Google Mock and the installation procedure changed, GTest integration was rewritten using the recommended `ExternalProject` method in CMake. New procedure ensures that GTest is built with the same compile flags as the testing project which uses it, at the expense of requiring a newer version of CMake.

_OpenCV_
The most problematic part was finding a prebuilt distribution of OpenCV; Trusty provides version 2.4.8, while our project requires ≥ 3.0.0. [Several PPAs](https://launchpad.net/ubuntu/+ppas?name_filter=opencv) providing OpenCV 3+ exist; one of them was chosen by trial and error.


gmrukwa:
[Enable Linux build](https://trello.com/c/Vyp9oQ0K/374-enable-linux-build)

gmrukwa (approved):
Just one quick question: will we be able at some point to keep definition of each lib separately, and then just merge them in this file?

Daroslaw (approved):

mg6:
@gmrukwa Sure, let’s address this in a separate pull request not to block our CI process further.

PR #51, opened 2018-06-04
Optimized ExpectationMaximization

Sand3r-:
Also additional enhancements for Matrix class, and optimized versions of Gaussian Distribution function
and class.


Sand3r-:
[Implement Optimized version of EM algorithm](https://trello.com/c/jYIGnC5r/497-implement-optimized-version-of-em-algorithm)

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
```diff
@@ -0,0 +1,113 @@
...
+                }
+            }
+            const unsigned lastK = numberOfComponents - 1;
+            for (unsigned i = 0; i < (unsigned)mzs.size(); i++)
+            {
+                DataType probability = m_Components[lastK].weight *
+                    Gaussian(mzs[i], m_Components[lastK].mean,
+                        m_Components[lastK].deviation);
+                m_AffilationMatrix[lastK][i] = probability;
+                m_Sums[i] = 1.0 / (m_Sums[i] + probability);
```
gmrukwa: Why does this differ from the version above? Could you add any comment or whatever?

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
```diff
@@ -0,0 +1,113 @@
...
+                {
+                    m_AffilationMatrix[k][i] *= m_Sums[i] * intensities[i];
+                }
+            }
+        }
+
+    private:
+        SpectrumView m_Spectrum;
+        Matrix<DataType> &m_AffilationMatrix;
+        std::vector<GaussianComponent> &m_Components;
```
gmrukwa: Who is the owner of this data? What if the owner gets deleted?

src/Spectre.libGaussianMixtureModelling/Matrix.h
```diff
@@ -31,29 +31,78 @@ struct Matrix
...
+    {
+        data = other.data;
+        other.data = nullptr;
+    }
+
+    /// <summary>
+    /// Move constructor
+    /// </summary>
+    /// <param name="other">Old matrix object.</param>
+    Matrix(Matrix& other)
```
gmrukwa: This isn't move constructor and definitely should not be here. It's better to delete this.

gmrukwa (commented):
What is the difference in benchmarks?

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
Sand3r-: The formula that is computed in this function is represented by:
image
Since the divison is a very costly operation, I decided to minimize it in such a way:
1. Compute the sum of probabilities (denominator) for each component from j=0 up to K-1 and save it in sums[i]
2. For the last component, add the final probability and take an inverse of the sums.
3. That way, in the loop below, we no longer have to perform division for each `K*N` elements, but we can apply a much more efficient multiplication. That way, we save ourselves (K-1)*N division operations.

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
Sand3r-: Owner of the data is the function using the class. I don't like this design either, but I had to implement it this way in order to conform with exisiting API which assumed storing the references to matrix and components upon creation. This will be refactored in the future.

src/Spectre.libGaussianMixtureModelling/Matrix.h
Sand3r-: I have no idea how did this end up here, thanks for catching it.

Sand3r-:
These changes have improved the efficiency of these functions 3 times, however it was the Expectation step that used to take so long.

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
```diff
@@ -19,14 +19,55 @@ limitations under the License.
...
+static inline void FilterLowHeightComponents(GaussianMixtureModel& components)
+{
+    constexpr DataType MIN_WEIGHT = 1.0e-3; // Height, weight are interchangable
+
+    GaussianMixtureModel::iterator pastEndIter =
+        std::remove_if(components.begin(), components.end(),
+            [MIN_WEIGHT](GaussianComponent& component)
+            { return component.weight < MIN_WEIGHT; });
+    size_t newSize = pastEndIter - components.begin();
+    components.resize(newSize);
```
Daroslaw, line 48: `resize` is better than `erase`?

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
```diff
@@ -19,14 +19,55 @@ limitations under the License.
...
+}
+
+static inline DataType CalculateDelta(GaussianMixtureModel& oldModel,
+    GaussianMixtureModel& newModel)
+{
+    DataType sumOfHeightDifferences = 0.0;
+    DataType sumOfScaledVarianceDifferences = 0.0;
+    DataType justASum = 0.0;
```
Daroslaw, line 56: xD

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
```diff
@@ -19,14 +19,55 @@ limitations under the License.
...
+    for (Index i = 0; i < oldModel.size(); i++)
+    {
+        justASum += newModel[i].weight;
+        sumOfHeightDifferences += abs(newModel[i].weight - oldModel[i].weight);
+        sumOfScaledVarianceDifferences +=
+            abs(newModel[i].deviation * newModel[i].deviation -
+                oldModel[i].deviation * oldModel[i].deviation) /
+               (newModel[i].deviation * newModel[i].deviation);
```
Daroslaw, line 64: I think you can squish some more performance by writing it as:
`1 - (oldModel[i].deviation * oldModel[i].deviation) / (newModel[i].deviation * newModel[i].deviation)`, because:
image
Or at least buffer the `newModel[i].deviation * newModel[i].deviation`.

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
Daroslaw: I think that passing these as arguments into `Expectation` function is a way to go.

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerOpt.h
```diff
@@ -0,0 +1,125 @@
...
+        /// Empty function made to conform with interface.
+        /// </summary>
+        void UpdateMeans()
+        {
+        }
+
+        /// <summary>
+        /// Empty function made to conform with interface.
+        /// </summary>
+        void UpdateStdDeviations()
```
Daroslaw: Are these functions meant to be filled in the future? If they are only to conform with interface why are they even called anywhere?

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerOpt.h
```diff
@@ -0,0 +1,125 @@
...
+        {
+        }
+
+    private:
+        SpectrumView m_Spectrum;
+        Matrix<DataType> &m_AffilationMatrix;
+        std::vector<GaussianComponent> &m_Components;
```
Daroslaw: Same as before

Daroslaw (dismissed):

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
Sand3r-: I checked the MSVC implementations, and noticed that resize has one less branch.

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
Sand3r-: Oh what xDDD cannot believe that I've actually left it that way xD

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerOpt.h
Sand3r-: Due to existing API this cannot be changed without refining everything else. 
To be refactored in the future.

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
Sand3r-: As we've discuessed on PM, that is not so easy to implement due to abs(),

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
gmrukwa: We are targeting multiplatform. Deployment will run clang / gcc.

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
gmrukwa: If design is bad, fix it right now. We do not want to merge poorly designed code. What is purpose of CR then? @Daroslaw is right in this case - this should be an argument into `Expectation` function or any other way around, not like this.

You cannot ensure from this point, that owner of this data does not get deleted. Situation you described is just one of possible usages. You do not prevent any other usage.

It's poorly designed, if correctness does not follow from design.

I'd accept it, when there is no other way. But there is.

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerOpt.h
gmrukwa: Such approach is bad. If API is unsuitable, make it fit. It's better to do it now, than to carry it longer, to let problem grow.

gmrukwa (changes requested):

src/Spectre.libGaussianMixtureModelling/GaussianDistribution.h
```diff
@@ -28,8 +28,48 @@ namespace spectre::unsupervised::gmm
...
+    /// </summary>
+    /// <param name="x">Argument to compute value for.</param>
+    DataType Compute(DataType x)
+    {
+        DataType diff = (x - m_Mean);
+        return m_Scalar * std::exp(m_Exponent * diff * diff);
+    }
+
+private:
+    DataType m_Mean;
```
gmrukwa: Aren't these `const`?

src/Spectre.libGaussianMixtureModelling/Matrix.h
gmrukwa: I've even meant to make `Matrix(Matrix& other) = delete;` and `Matrix(const Matrix& other) = delete;`. This should prevent us from making implicit copies.

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
Daroslaw: 😢

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
Sand3r-: Kinda hard to tell what's better at clang.
[clang - erase](https://github.com/llvm-mirror/libcxx/blob/master/include/vector#L1708)
[clang - resize](https://github.com/llvm-mirror/libcxx/blob/master/include/vector#L2037)
Still, both functions provide the same effect.

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
```diff
@@ -74,77 +74,69 @@ static inline DataType CalculateDelta(GaussianMixtureModel& oldModel,
...
+            FilterLowHeightComponents(components);
+            oldModel.assign(components.begin(), components.end());
+            m_Expectation.Expectation(m_AffilationMatrix, spectrum, components);
+            m_Maximization.Maximization(m_AffilationMatrix, spectrum,
+                components);
+            delta = CalculateDelta(oldModel, components);
         } while (delta > epsilon);
 
-        return m_Components; // Consider sorting before return
+        return components;
```
Daroslaw: If you modify the argument you don't need to return it. Entire function may be of type `void`

Daroslaw (dismissed):

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
Sand3r-: Yes I am aware of it, but to comply with the unwritten standard of native-algorithms libs, I decided to return it anyway, because otherwise @gmrukwa would complain.

Daroslaw (dismissed):

src/Spectre.libGaussianMixtureModelling/ExpectationMaximization.h
```diff
@@ -19,103 +19,125 @@ limitations under the License.
 */
 
 #pragma once
+#include <cmath>
 #include <random>
 #include "Spectre.libException/NullPointerException.h"
+#include "Spectre.libStatistics/Math.h"
 #include "DataTypes.h"
```
gmrukwa: I know this is not a part of this review, but please create task to make these paths w.r.t. solution root. Otherwise NuGet package won't work.

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerRef.h
```diff
@@ -40,60 +40,60 @@ limitations under the License.
...
-    /// <param name="components">Gaussian components.</param>
-    /// <exception cref="ArgumentNullException">Thrown when mzArray pointer are null</exception>
-    ExpectationRunnerRef(DataView mzs, Matrix<DataType> &affilationMatrix,
-                         std::vector<GaussianComponent> &components)
-        : m_Mzs(mzs), m_Components(components), m_AffilationMatrix(affilationMatrix)
+    /// <param name="spectrum">Spectrum from which array of m/z values is
+    /// fetched.</param>
+    ExpectationRunnerRef(SpectrumView spectrum)
     {
+        spectrum.mzs.size(); // Needed to avoid warning of unused param.
```
gmrukwa: It was usually solved by definition like `ExpectationRunnerRef(SpectrumView /*spectrum*/)`. This avoids strange, unnecessary call.

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
```diff
@@ -48,81 +48,114 @@ limitations under the License.
...
+    void UpdateWeights(Matrix<DataType> &affilationMatrix,
+        SpectrumView spectrum, GaussianMixtureModel& components)
     {
         // This part conducts the instruction:
         // "Using the gamma calculated in the Expectation step, 
         // calculate in the following order, for every k: (phi part)"
         // from the document
         DataType totalDataSize = 0.0;
-        for (unsigned i = 0; i < (unsigned)m_Spectrum.mzs.size(); i++)
+        for (unsigned i = 0; i < (unsigned)spectrum.mzs.size(); i++)
```
gmrukwa, line 100: Wait... Are you computing this each time you do _Maximization_ step? This should never change. What about precomputing this once and not repeating this operation?

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
```diff
@@ -48,81 +48,114 @@ limitations under the License.
...
+    /// of affilation of each sample to a certain gaussian component.
+    /// </param>
+    /// <param name="spectrum">Spectrum to base calculations on.</param>
     /// <param name="components">Gaussian components to be updated</param>
-    /// <exception cref="ArgumentNullException">Thrown when either of mzArray or intensities pointers are null</exception>
-    MaximizationRunnerRef(SpectrumView spectrum, Matrix<DataType> &affilationMatrix, std::vector<GaussianComponent> &components)
-        : m_Spectrum(spectrum), m_AffilationMatrix(affilationMatrix), m_Components(components)
+    void Maximization(Matrix<DataType> &affilationMatrix,
+        SpectrumView spectrum, GaussianMixtureModel& components)
     {
```
gmrukwa, line 77: Why do we have definitions in `.h` file?

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
```diff
@@ -48,81 +48,114 @@ limitations under the License.
...
-        const unsigned numberOfComponents = (unsigned)m_Components.size();
+        const unsigned numberOfComponents = (unsigned)components.size();
         for (unsigned k = 0; k < numberOfComponents; k++)
         {
             DataType weight = 0.0;
-            for (unsigned i = 0; i < (unsigned)m_Spectrum.mzs.size(); i++)
+            for (unsigned i = 0; i < (unsigned)spectrum.mzs.size(); i++)
             {
-                weight += m_AffilationMatrix[i][k] * m_Spectrum.intensities[i];
+                weight += affilationMatrix[k][i] * spectrum.intensities[i];
```
gmrukwa, line 111: Here is probably great underflow. You are summing like 200 000 floats in real data. Really small floats, coming from gaussian distribution. Check [this thread](https://stackoverflow.com/questions/2148149/how-to-sum-a-large-number-of-float-number).

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
```diff
@@ -48,81 +48,114 @@ limitations under the License.
...
         for (unsigned k = 0; k < numberOfComponents; k++)
         {
             DataType denominator = 0.0;
             DataType numerator = 0.0;
-            for (unsigned i = 0; i < (unsigned)m_Spectrum.mzs.size(); i++)
+            for (unsigned i = 0; i < (unsigned)spectrum.mzs.size(); i++)
             {
-                denominator += m_AffilationMatrix[i][k] * m_Spectrum.intensities[i];
-                numerator += m_AffilationMatrix[i][k] * m_Spectrum.mzs[i] * m_Spectrum.intensities[i];
+                denominator += affilationMatrix[k][i] * spectrum.intensities[i];
```
gmrukwa, line 139: Possible underflows.

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
```diff
@@ -133,24 +166,20 @@ class MaximizationRunnerRef
...
         {
             DataType denominator = 0.0;
             DataType numerator = 0.0;
-            for (unsigned i = 0; i < (unsigned)m_Spectrum.mzs.size(); i++)
+            for (unsigned i = 0; i < (unsigned)spectrum.mzs.size(); i++)
             {
-                denominator += m_AffilationMatrix[i][k] * m_Spectrum.intensities[i];
-                numerator += m_AffilationMatrix[i][k] *
-                    pow(m_Spectrum.mzs[i] - m_Components[k].mean, 2) * m_Spectrum.intensities[i];
+                denominator += affilationMatrix[k][i] * spectrum.intensities[i];
```
gmrukwa, line 175: Possible underflows.

src/Spectre.libGaussianMixtureModelling/Matrix.h
```diff
@@ -31,29 +31,68 @@ struct Matrix
...
-            data[0] = new Type[width * height]; // allocate data once
+            data[0] = new Type[length]; // allocate data once
             for (unsigned i = 1; i < height; ++i)
                 data[i] = data[0] + i * width; // assign pointers
         }
-        memset(data[0], 0, width * height * sizeof(Type));
+        memset(data[0], 0, length * sizeof(Type));
     }
 
+    /// <summary>
```
gmrukwa, line 51: If this class gets extended even a bit more, switch to `cv::Mat` or just consider using Eigen.

src/Spectre.libGaussianMixtureModelling/Matrix.h
```diff
@@ -66,6 +105,22 @@ struct Matrix
...
+    /// Checks if two matrices are equal.
+    /// </summary>
+    /// <param name="other">Matrix to compare with.</param>
+    /// <param name="length">Width*height of the matrices</param>
+    /// <returns>True if matrices are equal, false otherwise</returns>
+    bool Equals(const Matrix& other, unsigned length)
+    {
+        for (unsigned i = 0; i < length; i++)
+        {
+            if (data[0][i] != other.data[0][i])
```
gmrukwa, line 118: We are using this for `float`s mostly. What about epsilon-based comparison?

src/Spectre.libGaussianMixtureModelling/GaussianDistribution.h
```diff
@@ -28,8 +28,47 @@ namespace spectre::unsupervised::gmm
...
+class GaussianDistribution
+{
+static constexpr DataType invSqrtOfTwoPi = 0.3989422804014327; // 1/sqrt(2pi)
+public:
+    /// <summary>
+    /// Constructs a distrubution based on given mean and standard deviation.
+    /// </summary>
+    /// <param name="mean">Mean of the distribution.</param>
+    /// <param name="std">Standard deviation of the distribution.</param>
+    GaussianDistribution(DataType mean, DataType std) : m_Mean(mean),
```
gmrukwa, line 53: We will need information on `std` and _weight_ in final model - this is used for components merging, etc. How do you preserve this? Do you return this at some point other way?

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerOpt.h
```diff
@@ -0,0 +1,106 @@
...
+            const unsigned numberOfComponents = (unsigned)components.size();
+            const unsigned dataSize = (unsigned)spectrum.mzs.size();
+            DataView mzs = spectrum.mzs;
+            for (unsigned k = 0; k < numberOfComponents; k++)
+            {
+                DataType probabilitySum = 0.0;
+                DataType meanNumerator = 0.0;
+                for (unsigned i = 0; i < (unsigned)mzs.size(); i++)
+                {
+                    probabilitySum += affilationMatrix[k][i];
```
gmrukwa: Possible underflows.

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerOpt.h
```diff
@@ -0,0 +1,106 @@
...
+                    meanNumerator += affilationMatrix[k][i] * mzs[i];
+                }
+                components[k].weight = probabilitySum / m_SummedIntensities;
+                components[k].mean = meanNumerator / probabilitySum;
+                DataType stdNumerator = 0.0;
+                for (unsigned i = 0; i < (unsigned)mzs.size(); i++)
+                {
+                    DataType diff = mzs[i] - components[k].mean;
+                    stdNumerator += affilationMatrix[k][i] * diff * diff;
```
gmrukwa: Possible underflows.

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
```diff
@@ -0,0 +1,105 @@
...
+        const unsigned numberOfComponents = (unsigned)components.size();
+        for (unsigned k = 0; k < numberOfComponents; k++)
+        {
+            for (unsigned i = 0; i < (unsigned)mzs.size(); i++)
+            {
+                DataType probability = components[k].weight *
+                    Gaussian(mzs[i], components[k].mean,
+                        components[k].deviation);
+                affilationMatrix[k][i] = probability;
+                m_Sums[i] += probability;
```
gmrukwa: Possible underflows.

gmrukwa (commented):
Please explain, how did you approach underflow problem in the EM for real data we will be processing (200k m/z-s). When this gets clear, I click *Approve*.

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
Sand3r-: It is precomputed in the optimized version MaximizationRunnerOpt.h

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
Sand3r-: That's how the previous API used to look like. It limits the number of translation units and allows compiler for doing better optimizations.

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
Sand3r-: This is a reference version which won't be likely to be used.

src/Spectre.libGaussianMixtureModelling/Matrix.h
Sand3r-: That was made just for the purpose of tests with indices (yet to come in Dyn Prog initialization)). I'd need to create a template specialized method for it.

src/Spectre.libGaussianMixtureModelling/GaussianDistribution.h
Sand3r-: These are saved in Gaussian Components.

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
```diff
@@ -73,13 +73,19 @@ class ExpectationRunnerOpt
         const unsigned numberOfComponents = (unsigned)components.size();
         for (unsigned k = 0; k < numberOfComponents; k++)
         {
+            // Using Kahan summation algorithm for improved accuracy
```
gmrukwa: Can we extract this into a reusable element?

gmrukwa (approved):
Seems OK now :) Good job!

src/Spectre.libGaussianMixtureModelling/MaximizationRunnerRef.h
gmrukwa: This requires recompilation in each dependent project, while step itself will be used in one context: context of GMM. I'd understand such approach with your argument for GMM itself, but not for its elements. Maybe you could clarify it a bit more?

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
Sand3r-: I've considered doing that, but in another PR.

Daroslaw (approved):

src/Spectre.libGaussianMixtureModelling/ExpectationRunnerOpt.h
gmrukwa: Add a task then.

PR #52, opened 2018-06-04
Provide Dockerfiles for local build testing with GCC/Clang

mg6:
Please consult contrib/docker-build/README.md for usage info.

Added .dockerignore file to globally exclude build artifact directories
when attempting an image build. This heavily affects output image size.

Dockerfile build instructions are divided into two RUN commands: for
dependencies and proper build script; this has a possitive effect on
rebuild time, as the first RUN creates an intermediate container with
all required dependencies, while the second RUN is always refreshed
with new files from repository, before a C++ build is attempted.

### DIFF
--- a/contrib/docker-build/README.md
+++ b/contrib/docker-build/README.md
@@ -1,0 +1,47 @@
+# Docker-based build images
+
+Meant for local testing, before sending changes to Travis CI.
+
+## Usage
+
+Run commands from repository root. Successful image build acts as a smoke test.
+
+Local installation of Docker is required.
+
+### GCC
+
+```sh
+docker build \
+  -f contrib/docker-build/gcc.Dockerfile \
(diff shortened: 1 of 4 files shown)


contrib/docker-build/README.md
```diff
@@ -0,0 +1,45 @@
+# Docker-based build images
+
+Meant for local testing, before sending changes to Travis CI. Require local installation of Docker.
```
Sand3r-: Requires*</grammar_nazi>

gmrukwa (dismissed):

Sand3r- (dismissed):

contrib/docker-build/README.md
mg6: Redacted and rebased.

Sand3r- (approved):

gmrukwa (approved):

PR #53, opened 2018-06-17
[WIP] Implement accurate sum

Sand3r-:
**Work in progress, low priority.**
These commits implement an accurate summing algorithm using combined:
* [Kahan summation algorithm](https://en.wikipedia.org/wiki/Kahan_summation_algorithm)
* Sum-dispersed approach to summing

The latter employs the idea of several sums (in case of ymm registers, 8 sums) to be computed simultaneously, so that in the end, their magnitudes are similar and they can be summed without losing much of a precision.

The tests have shown, that in the case of using std::accumulate for summing an arithmetic series starting at 1 going all the way up to 1 mln with a step 1, the final error is equal to 59146241.
In the meantime, using the AccurateSum not only speeds up the summing by 3 times, but decreases the final error to the magnitude of 0.0000001. In other words, it is 591 trillion times more accurate than standard accumulate. That is of course caused by the fact that in std::acumulate at certain point of adding numbers that largely vary in magnitude (single entity + sum) the sum is not changed due to lack of floating-point precision.

We'll also find out whether Travis employs AVX2 capable processors.


Sand3r-:
Temp closed until building on several HW configurations is supported.

PR #55, opened 2018-07-02
Splitter Segment Decomposer

Sand3r-:
Implements Splitter Segment decomposition which does the following:
- Initialization during which many later needed parameters are evaluated

In the loop:
- Picks a certain amount of gaussian components
- Initializes it with Dynamic Programming
- Estimates signal with Expectation Maximization
- Computes the Scale and Quality of the approximation
- Based on amount of components and given quality it either stops, or increments number of components and returns to the beginning of the loop

Note to the last point: The less components are used, the better.
The result of the computations is a set of Gaussian Curve parameters that approximate the Splitter Segment Signal.
After the loop:
- From among the resulting Gaussian Components, algorithm chooses those that approximate the Splitter 
- If there are any components between the leftmost and rightmost splitter-building-component, they are automatically included into the set of gaussian approximating the splitter.
- Gaussian components building up the splitter are returned

Sorry, I did not have time to make this much more comprehensible. If you have any questions, please head them directly to me, and I'll be glad to explain.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ environment:
 configuration:
   - Debug
   - Release
-version: 4.7.0.{build}
+version: 4.8.0.{build}
 
 init:
 - cmd: echo Project - %APPVEYOR_PROJECT_NAME%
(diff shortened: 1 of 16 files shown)


Sand3r-:
[Implement Splitter Segment Decomposer](https://trello.com/c/gLHHF4DH/514-implement-splitter-segment-decomposer)

Regis128 (dismissed):
seems ok

Daroslaw (approved):

Regis128 (approved):

